PR #116, opened 2020-01-03
Adds new plugin [ikaruswill/lovelace-fan-xiaomi]

ikaruswill:
Before you submit a pull request, please make sure you have the following:

- [x] You have tested it with HACS by adding it as a custom repository

### DIFF
--- a/plugin
+++ b/plugin
@@ -81,5 +81,6 @@
   "xMrVizzy/button-toolbar",
   "fineemb/lovelace-fan-xiaomi",
   "Anonym-tsk/lovelace-starline-card",
-  "dcramer/lovelace-nextbus-card"
+  "dcramer/lovelace-nextbus-card",
+  "ikaruswill/lovelace-fan-xiaomi"
 ]


ikaruswill:
Hey guys, Happy New Year!

I understand this is a forked repository and I know the first thing that comes to mind is, why not submit a PR against the original repository?

That was indeed on my mind. But I've decided not to.

Here're my thoughts that made me submit this as a separate plugin, correct me if I'm wrong:
* The original repository is in Chinese
    * There was an issue request filed for translation but I believe neither me nor the original author has the technical capabilities for a settings page and language toggle. This is meant to be simple.
* The original plugin name in hacs.json is in Chinese `小米电风扇`
    * Mine here `lovelace-fan-xiaomi`. 
    * There is no namespace conflict nor target audience conflict/confusion. Chinese people will find the Chinese version, English speaking people will find the English version.
* The original feature set works best for older-model devices. 
    * New devices are missing temperature, humidity, battery sensors
    * I have made significant opinionated modifications to the feature-set so that it works well with the new v4 device, and all past models.

Due to the above reasons, a PR will not make much sense as it pretty much overwrites the original feature-set and intended target-audience.

With that, I seek your approval for this PR. Cheers.

ludeeus:
It's better to have it as a fork as you have here, then to just copy the content and republish (I have seen that a few times too.)

As for translations, I recently looked into that, and it was easier then I imagined <https://github.com/custom-cards/boilerplate-card/pull/16/commits/19a88abd43fdfbcccc1244723a08d9c387c9db23>

It would be preferable not having 2 elements for the same thing, with that said since all instructions of the other repo is in Chinese I'll allow this one, and will add some more "rules" for publishing later. (Instructions should be in English or a country definition must be present in hacs.json)

ikaruswill:
@ludeeus Appreciate that.
That's an interesting find on translations, I'll look into it.

PR #290, opened 2020-03-24
Adds new integration [ollo69/ha-samsungtv-smart]

ollo69:


ludeeus:
What does this provides that the 3 other samsung integrations does not, why is 4 different integrations needed?

ollo69:
First of all this component can be configured using config_flow.
It use SmartThings, but compared to the original from which I forked, configuration is mutch simpler . I also rewrite many part of the code to use async. I initially PR my change, but I realize that I did to mutch change compared to the master.

PR #349, opened 2020-04-22
Adds new plugin [maxwroc/battery-state-card]

maxwroc:


azogue:
Hi @maxwroc 

I'm trying to add the repo as custom repo, using `https://github.com/maxwroc/battery-state-card` + "Plugin" but it doesn't work.

Am I missing something?

Error alert is:

```
An error occurred while processing
Exception: HacsException
Message: Repostitory structure for 0.9.0 is not compliant
Could not add this repository, make sure it is compliant with HACS.
```

maxwroc:
@azogue Thanks for letting me know. It was working yesterday. I will check what has happened. The only thing which I have changed was that I have added more files to latest release as the PR check was not passing without the js file attached to release (even though doc states that zip release is supported and you just need to add `zip_release: true` in your hacs.json)

azogue:
> @azogue Thanks for letting me know. It was working yesterday. I will check what has happened. The only thing which I have changed was that I have added more files to latest release as the PR check was not passing without the js file attached to release (even though doc states that zip release is supported and you just need to add `zip_release: true` in your hacs.json)

Maybe the only thing you need is a new release tag, pointing to the last commit?

ludeeus:
The bot is not smart enough to check for `zip_release`.
I'm working on a new implementation of the checks done in PR's against this repo (hopefully it will be the exact same code as HACS itself).

maxwroc:
Ok that was it! I have deleted rest of the files from release and I have left just a zip file. Now it works and you can add it to HACS. But probably if I rerun the PR checks here they won't pass :/ 

It would be good if one of the owners/contributors take a look here. I think it should be working fine for people after merging this PR. It is just the PR check which is faulty and does not handle correctly zip_release option.

maxwroc:
HACS guys one more thing. When you have a zip_release there is invalid info added on the bottom of the README (after installation)

image

It would be good to add ability to override this value via hacs.json or actually fix it (as the name of the main file is actually enforced by the rules so you know it)

ludeeus:
Actually...
I'm not sure that zip_release is supported for other than integrations.
I'll have a look at the code after work.

azogue:
I can confirm the local url ending in `.zip` is also copied to resources, so unless a manual edit of that extension to `.js`, install doesn't work.

BTW, @maxwroc, I cannot access the more-info for the battery sensors, is that expected?

azogue:
> install doesn't work.

I mean, install works, but lovelace resources not. Just with the extension change it works

maxwroc:
Are you talking about the case when yaml config is disabled, right? If so I haven't tested it TBH - I was adding entry manually in lovelace ui config.

azogue:
> You're talking about the case when yaml config is disabled, right? If so I haven't tested it TBH - I was adding entry manually in lovelace ui config.

Yes. In fact, that is the flow to test, as we want things to install smoothly for all users, don't you think? ;-)

Also, in modern HA, dashboards (there are more than one now) are separated from resources (which are shared) and have their own place in the menus

ludeeus:
It does not matter, they do the same thing.

maxwroc:
Ok I see now that it doesn't allow to add repo when release is with more then one js file but map files are fine and they are being downloaded as well.

maxwroc:
Closing and reopening PR to retrigger checks after last changes made on repo/releases

ludeeus:
I'm almost done rewriting the checks.
Nothing will be merged until that is done.

ludeeus:
https://github.com/hacs/default/runs/612662752?check_suite_focus=true#step:6:13
Now PRs is checked with HACS.
There are still a few things missing from that action before I can switch over from the bot.

maxwroc:
Thanks @ludeeus

PR #391, opened 2020-05-02
Adds new plugin [fineemb/lovelace-air-filter-card]

fineemb:

### DIFF
--- a/plugin
+++ b/plugin
@@ -28,6 +28,7 @@
   "dooz127/swipe-glance-card",
   "enriqg9/dual-thermostat",
   "faeibson/lovelace-multiline-text-input-card",
+  "fineemb/lovelace-air-filter-card",
   "fineemb/lovelace-cn-map-card",
   "fineemb/lovelace-dc1-card",
   "fineemb/lovelace-fan-xiaomi",


fineemb:
> is this Xiaomi specific or a generic Air Filter (Purifier) Card ?

This is based on Xiaomi

fineemb:
Currently only tested on Xiaomi devices, other devices are uncertain

fineemb:
It is not very difficult to apply to other devices, but I have not tested the environment
Maybe I should explain it in the README

GeorgeSG:
Not sure why I was tagged as a reviewer so I won't be approving/rejecting this, but I'd add my two cents.

The card looks great and it'd be awesome if you can make it a bit more generic so it can work with more devices!

I don't have a Xiaomi filter, but I have a Philips one ([integration](https://github.com/xMrVizzy/philips-airpurifier)). I tried it and it doesn't work right now, but it could with some small changes. Here are my suggestions:
1. The card breaks if you pass less than 3 `modes`. This is not documented, but ideally, it should work with any number of modes. IMO This is the most important fix, because I think it'll break even with Xiaomi devices.
2. You can make the `aqi` property configurable. My purifier has a `pm25` attribute. Adding a config option for `aqi_attribute` that defaults to `aqi` will enable more devices to use this card
3. Same for speeds/modes - it has hardcoded ` this.modes = ["auto","silent","strong"]` and I don't think it'll use others, even if you pass them as config. For example, my Filter's auto mode value is "Auto Mode".

ludeeus:
> Not sure why I was tagged as a reviewer so I won't be approving/rejecting this, but I'll add my two cents.

@GeorgeSG This is why (which is just rude):
image

PR #449, opened 2020-05-16
Adds new integration [hasl-sensor/integration]

DSorlov:
HASL integration is moving from DSorlov to hasl-sensor and renaming to integration instead of hasl-platform

- [X] You are submitting only 1 repository.
- [X] You repository is compliant with https://hacs.xyz/docs/publish/start
- [X] You have tested it with HACS by adding it as a custom repository.
- [X] The list are still alphabetical after my change.


DSorlov:
Unsure of why it halts on homeassistant in hacs.json? According to docs this is a valid setting, and is same in the currently active repo?

ludeeus:
Why did you do this instead of just transferring it?
Now I have to blacklist the old repo...

ludeeus:
> Unsure of why it halts on homeassistant in hacs.json?

hacs.json is valid for that, manifest.json is not ;)

DSorlov:
Ah! Did not think about that. Dang. The reason was mostly because the new version will have a completely new motor and Im not conviced that all would agree. Perhaps I should just launch the nwe version in a new domain instead depending on flavor. hacs3 instead of hacs. I will close this for now and think abit about it.. thanks for the feedback and the subtle hint on wrong file.. ;P

DSorlov:
@ludeeus Just realized that you already moved hasl to blacklist?
HASL at http://github.com/DSorlov/hasl-platform will continue
HASL3 at http://github.com/hasl-sensor/integration is a new branch NOT replacing the old...

PR #568, opened 2020-08-06
Adds new integration [FL550/dwd_weather]

FL550:


ludeeus:
Is this integration limited to a single (or a few countries)?

FL550:
It's available worldwide, but outside Europe and especially Germany the available stations are limited.

PR #579, opened 2020-08-14
Adds new plugin [finity69x2/cover-position-preset-row]

finity69x2:


finity69x2:
Sorry I have no idea why the "initialize" check is failing

PR #584, opened 2020-08-18
Adds new plugin [ezand/lovelace-posten-card]

ezand:
Adding a Lovelace card that displays Norwegian mail delivery days.


ludeeus:
This only applies to a single country, add the country key in the hacs.json file

ezand:
Ah of course, it's added to `hacs.json` now
image

PR #593, opened 2020-08-27
Adds new integration [KoljaWindeler/kaco]

KoljaWindeler:
added two further custom_components for solar power based sensors


KoljaWindeler:
hmm how can I retrigger the check?

PR #629, opened 2020-09-23
Adds new plugin [turbulator/pandora-cas-card]

turbulator:


JonahKr (approved):
Generally LGTM 
@turbulator i cannot chack if the images are licensed, especially as they are just base64 encoded png raw files which are hardcoded.

ludeeus (changes requested):
README is not in English, which makes it only useful to a part of the world, add the country key to the hacs.json file as you did for the integration

turbulator:
You are absolutely right. I've just added country section in hacs.json.

PR #653, opened 2020-10-13
Adds new integration [segalion/securitasdirect]

segalion:
I am trying to include https://github.com/segalion/securitasdirect in HACS.

Thanks in advance.


segalion:
Fail. Try second time

PR #670, opened 2020-10-21
Adds new plugin [htmltiger/numberbox-card]

htmltiger:
Please add this
[https://community.home-assistant.io/t/addon-input-number-with-and-buttons-to-replace-slider](https://community.home-assistant.io/t/addon-input-number-with-and-buttons-to-replace-slider)

### DIFF
--- a/plugin
+++ b/plugin
@@ -93,6 +93,7 @@
   "gurbyz/power-wheel-card",
   "hasl-sensor/lovelace-hasl-departure-card",
   "hasl-sensor/lovelace-hasl-traffic-status-card",
+  "htmltiger/numberbox-card",
   "iantrich/aftership-card",
   "iantrich/config-template-card",
   "iantrich/podcast-card",


KTibow:
Your link is broken

maxwroc:
I'm signing off this PR as adding your repo via hacs and installation was successful. Although I have found few issues:

* There was a wrong path in the lovelace configuration in your README (I've sent PR with the fix).

* Looks like the name doesn't show up by default (when there is no `name` setting)

* Name setting seems to be not respected at all - setting it to `true` doesn't help nor putting a string there

* Clicked/tapped button stays highlighted until you click somewhere else 

image

maxwroc (approved):

htmltiger:
> I'm signing off this PR as adding your repo via hacs and installation was successful. Although I have found few issues:
> 
> * There was a wrong path in the lovelace configuration in your README (I've sent PR with the fix).
> * Looks like the name doesn't show up by default (when there is no `name` setting)
> * Name setting seems to be not respected at all - setting it to `true` doesn't help nor putting a string there
> * Clicked/tapped button stays highlighted until you click somewhere else
> 

The name works fine here (Home Assistant 0.116.4)
https://imgur.com/N52LKv9
Will need to work on button focus blur.

Thanks

KTibow:
@htmltiger they meant that it should grab the `friendly_name` from the entity, so you don't have to manually put a name in config.

htmltiger:
> @htmltiger they meant that it should grab the `friendly_name` from the entity, so you don't have to manually put a name in config.

it does that already
https://imgur.com/D89cJTG

KTibow:
Aaah, got it. Thanks for clarifying.

maxwroc:
It wasn't working for me. In the example I have pasted the name wasn't showing at all regardless if name property was specified, was true/false or set to done text.

I'm on 116.4 too and tested it on Chrome PC/Android

maxwroc:
Ok I have taken a look what is going on. You have the condition where you don't render name if there is no icon. Since I haven't set icon explicitly and it wasn't available via attributes nothing was showing up except the +/- and value.
These are the attributes of the entity:
image

One more thing which is just an advice/suggestion:  your card is being re-rendered on every hass update - which happens quite often. The reason is that _hass and stateObj are on the list of tracked properties and they are being set every time. Even though these objects might be the same as previous (same properties and values) they are not the same instance. So when LitElement is doing simple comparison "==" to check whether update is required it always re-render it. 
To make it working correctly you have two options:
* Implement shouldUpdate function where you will be checking if the actual update is needed
* Remove hass and stateObj from tracked properties and use only simple properties as tracked ones (like strings/numbers/bools)

IMO the second approach is better. Anyway this is just a hint - if you want to make it better

htmltiger:
> Ok I have taken a look what is going on. You have the condition where you don't render name if there is no icon. Since I haven't set icon explicitly and it wasn't available via attributes so nothing was showing up except the +/- and value.
> These are the attributes of the entity:
> 
> One more thing which is just an advice/suggestion: your card is being re-rendered on every hass update - which happens quite often. The reason is that _hass and stateObj are on the list of tracked properties and they are being set every time. Even though these objects might be the same as previous (same properties and values) they are not the same instance. So when LitElement is doing simple comparison "==" to check whether update is required it always re-render it.
> To make it working correctly you have two options:
> 
> * Implement shouldUpdate function where you will be checking if the actual update is needed
> * Remove hass and stateObj from tracked properties and use only simple properties as tracked ones (like strings/numbers/bools)
> 
> IMO the second approach is better. Anyway this is just a hint - if you want to make it better

I have updated it to display name even if icon is none, 
also added shouldUpdate function so now it only renders if changed.
Thanks for the feedback.

PR #671, opened 2020-10-22
Adds new integration [basnijholt/adaptive-lighting]

basnijholt:
Brands PR https://github.com/home-assistant/brands/pull/1900


basnijholt:
@ludeeus, the brands PR does actually exist at https://github.com/home-assistant/brands/pull/1900

Otherwise, the CI should pass now.

ludeeus:
That PR is still open.
You already have a PR open for core, I'm not sure this should be added here as well, only to be removed later.
That will cause a lot of questions/issues when it's removed, and even if it's not in here, it can still be used for testing

basnijholt:
> You already have a PR open for core, I'm not sure this should be added here as well, only to be removed later.

That is true, but that PR has been open for about three months and hasn't received any review from a core dev. I can imagine it will take months more before it is merged. This gives people a good way to try it and gives me the opportunity to squeeze out all the bugs before it makes it to core :)

ludeeus:
Like I said, it can still be used without the addition here?
And without a merged brands PR, it will not be accepted here.

PR #679, opened 2020-10-27
Adds new integration [Sennevds/media_player.template]

Sennevds:


KTibow:
Create a new release and ask @hacs-bot to recheck.

Sennevds:
/recheck
Aha didn't know it checked the release zip instead of the repo itself

Sennevds:
If I add template as as a dependecy hass won't load my integration because template isn't a complete integration. If I leave it out hacs check fails what should I do with this?

ludeeus:
Adding to  `after_dependencies` should be safe

Sennevds:
Thanks for the fast response. That was the problem

Sennevds:
closes Sennevds/media_player.template#1

PR #697, opened 2020-11-08
Adds new integration [asev/homeassistant-helios]

asev:


ludeeus (changes requested):
<https://github.com/asev/homeassistant-helios/blob/master/hacs.json#L4>

Needs to be `true`, not `"True"`

ludeeus (approved):

PR #709, opened 2020-11-21
Adds new integration [andker87/ialarm]

andker87:
Added ialarm integration - custom component for Home Assistant alarm


RyuzakiKK:
FYI I updated the pyialarm library and now it uses the iAlarm API instead of webscraping the config page.

The iAlarm intergration should be included again in HA core https://github.com/home-assistant/core/pull/43525

ludeeus:
With this custom component pinning homeassistant as a requirement (which will cause a log of issues for anyone trying to use it) <https://github.com/andker87/ialarm/blob/master/custom_components/ialarm/manifest.json#L9>, and since there is an active PR to add this integration back to core, I'm closing this PR.

PR #731, opened 2020-12-08
Adds new integration [Limych/ha-car_winter_tire]

Limych:


integration
```diff
@@ -256,6 +256,7 @@
   "Limych/ha-average",
   "Limych/ha-beward",
   "Limych/ha-car_wash",
+  "Limych/ha-car_winter_tire",
```
KTibow, line 259:
```suggestion
  "Limych/ha-car_winter_tires",
```

KTibow: Mind making that tweak? Right now this just makes a 404.

PR #738, opened 2020-12-13
Adds new plugin [ben8p/lovelace-auto-reload-card]

ben8p:
Custom [Lovelace](https://www.home-assistant.io/lovelace) card to use in [Home assistant](https://www.home-assistant.io/) allowing lovelace to auto reload every X minutes.

When running on wall tablet, it often happens that lovelace websocket disconnect and the entities are not refreshed automatically anymore.  
This card, can be used to workaround this problem.  
Hopefully, one day, this card will be deprecated and a real fix will be available.


KTibow:
You can disable websocket auto-disconnect, if it might be malfunctioning:
image

ben8p:
Cool!
I never heard of this setting before.

I will give it a try

ben8p:
> You can disable websocket auto-disconnect, if it might be malfunctioning:
> image

Replying here as it might help someone:
I tried disabling the socket disconnection, this resulted is hassio crashing every 24 hours or so.

Using this refresh card does not crash hassio (but it comes with other drawback like ui might refresh while you are using it)

KTibow:
You could just detect whether there's any interaction in the past 5 minutes.

PR #741, opened 2020-12-17
Adds new plugin [sopelj/lovelace-kanji-clock-card]

sopelj:
A card for Lovelace that displays a clock using Japanese Kanji for date and time. 
I know it's a bit niche, so I hope it's ok. I just thought others might enjoy it too.


ludeeus (changes requested):
Your hacs.json file needs a `country` key https://hacs.xyz/docs/publish/start#hacsjson

sopelj:
For sure! Sorry, I thought it was optional. 
This may be a dumb question, but what country? Mine (CA)? Or should I put Japan (JP).
I thought this may just have been for region locked integrations/entities.

ludeeus:
jp is probably most fitting here, since it's in that language

sopelj:
OK, that makes sense. Thank you!
I've added it

PR #748, opened 2020-12-19
Adds new integration [dgomes/rrd_recorder]

dgomes:
Adds new integration [dgomes/rrd_recorder] while still waiting for inclusion to Core


dgomes:
This PR requires the user to install "rrdtool-dev" library, can't build the wheel without it... so I can't get this through HA...

dgomes:
:(

ludeeus:
When/if you get the needed wheels (and the other checks) sorted out, we can revisit this.
But until then, it does not make sense to distribute it.

dgomes:
Users using core only are using it (they can simply install the C library dependency), I just haven't figured out the HA way of expressing that need.

I sure hope that it doesn't require that the Dockerfile needs to be changed

ludeeus:
If it's only needed to build the wheel, and not during runtime, adding it to the list of apk's in the builder in the wheels repo for custom integrations should be enough.

dgomes:
I did that, but it fails to validate because it doesn't have the C library in the validation container

https://github.com/home-assistant/wheels-custom-integrations/pull/193

ludeeus:
Do a rebase, a lot has changed in that repo since you opened that PR

dgomes:
It's always a catch22 ... I need to add the dependency to the GitHub workflows before it can actually run an validate the PR ...

ludeeus:
I wonder if that job could/should be migrated to a container :thinking:

dgomes:
I don't understand how other requirements don't face the same issue, or what am I doing wrong 😕

ludeeus:
That would be nice, but currently, that's not an option

PR #778, opened 2021-01-14
Adds new integration [fondberg/easee_hass]

tmjo:
Add Easee EV charger integration


tmjo:
Failing on wheel (error code 404) but as far as I can see, things are good in the wheel:

https://github.com/home-assistant/wheels-custom-integrations/pull/243

PR #780, opened 2021-01-17
Adds new integration [jtbgroup/kodi-media-sensors]

jtbgroup:
I'd like my new integration component to be available via HACS (jtbgroup/kodi-media-sensors).


jtbgroup:
Removed the requirement  in the manifest.json because there is no dependency. Now check passes. Only the brand logo are not approved yet.

PR #788, opened 2021-01-22
Adds new integration [marotoweb/home-assistant-vacuum-viomise]

marotoweb:
New integration for Viomi SE vacuum support


ludeeus (changes requested):
https://github.com/hacs/default/runs/1857885373?check_suite_focus=true#step:2:408

PR #818, opened 2021-02-12
Adds new integration [Antoni-Czaplicki/vulcan-for-hassio]

Antoni-Czaplicki:


ludeeus (changes requested):
The library you use installs `asyncio` from PyPi, which it should not, that is a standard lib in Python.

Antoni-Czaplicki:
@ludeeus is there any other option to change this other than removing this dependency from the API?

ludeeus:
You have to take that up with the owner of that library

ludeeus (changes requested):
This service is only available in a single country, add the country key to hacs.json

ludeeus (approved):

PR #825, opened 2021-02-16
Adds new integration [mawinkler/astroweather]

mawinkler:


ludeeus:
Your `setup` logs should really be `debug` and not `info`

PR #840, opened 2021-02-26
Adds new integration [nick2525/broadlink_s1c_s2c]

nick2525:


nick2525:
@ludeeus It is very strange that I should also commit to the official repository of home-assistant (brands) Ok, I tried to create a pull request there, https://github.com/home-assistant/brands/pull/2293 but there is a bug with CLA signing https://github.com/home-assistant/brands/issues/2294  so I can't add icons(

PR #874, opened 2021-03-20
Adds new integration [alandtse/pr_custom_component]

alandtse:


dreed47 (approved):

ludeeus (changes requested):
<https://github.com/alandtse/pr_custom_component/blob/a5f55d0828f2567dfe399fd75f4b7542010f81da/custom_components/pr_custom_component/__init__.py#L112-L126>

I cannot allow this to be included as default with this in there, each API client should use their own API key.

PR #912, opened 2021-04-19
Adds new plugin [estevez-dev/extended-banner-card]

estevez-dev:


ludeeus:
Why do a separate card instead of trying to add functionality to the existing banner card?

estevez-dev:
> Why do a separate card instead of trying to add functionality to the existing banner card?

I was trying to do so, but [boilerplate-card](https://github.com/custom-cards/boilerplate-card) has so many things already implemented that it was much easier to migrate existing banner-card functionality to the boilerplate-cars codebase.

PR #914, opened 2021-04-21
Adds new integration [thomasgermain/vaillant-component]

thomasgermain:


dreed47 (approved):

ludeeus (changes requested):
CI does not pass

ludeeus:
This should not be added as a default <https://github.com/thomasgermain/vaillant-component/issues/26#issuecomment-726195087>
Since there are plans to have it as core, it will need to be removed later, which does not make much sense (I will update the docs later with that)
Users can still use it as a custom integration.

thomasgermain:
Well yeah, the end goal is to have it included in core, but I don't know if it will be the case one day. 
The goal with default HACS inclusion is to be sure that the integration is quite clean (from code point of view) and maybe have more visibility.

If you think I should not include it in HACS, then I'm perfectly ok with custom repo option

ludeeus:
Offering it as a default, and then having to remove it later will cause questions/issues for everyone that use it, that will not be a great experience for the user.

So while HACS is great for alpha/beta test integrations before adding it to core, I think when that is done it should be done as custom and not included as default.

thomasgermain:
Sounds good to me, thanks for your input. I'll close the PR. 

Have a nice day :)

PR #916, opened 2021-04-24
Adds new integration [koying/openrgb_ha]

koying:


ludeeus (changes requested):
CI does not pass

koying:
@ludeeus  Should be fixed

PR #926, opened 2021-05-03
Adds new plugin [JurajNyiri/PlexMeetsHomeAssistant]

JurajNyiri:
Adds new plugin [JurajNyiri/PlexMeetsHomeAssistant]


dreed47 (approved):

ludeeus (changes requested):
This is not an integration, remove all references to calling it one

JurajNyiri:
Good catch.

Removed all mentions of integration and replaced with card to make it clear its a card.

PR #929, opened 2021-05-08
Adds new integration [benleb/surepetcare]

benleb:


benleb:
🙄 
image

ludeeus (changes requested):
CI does not pass

benleb:
https://github.com/home-assistant/wheels-custom-integrations/pull/404

it fails because I had the change the integration domain `surepetcarebeta`  and repo name `surepetcare` do not match? 😂

ludeeus:
repo name does not matter

benleb:
but that of the json file? 😂 I think that would be something I can fix...^^

ludeeus:
HACS (and HA itself) uses the `domain` key of the manifest.json file for integrations.

benleb:
ah okay nice 👍 that means it's the workflow which is buggy ;) https://github.com/benleb/surepetcare/blob/dev/manifest.json

ludeeus:
it's not

benleb:
any news?

gjohansson-ST:
Does this makes sense to get into HACS even if it would comply with CI?
It clearly states it's development, readme is not correct and manifest point to HA documentation.

benleb:
> Does this makes sense to get into HACS even if it would comply with CI?
> It clearly states it's development, readme is not correct and manifest point to HA documentation.
> Why not making a PR to the official integration instead?

you know that I am the developer of the official integration?^^ you know why this exists at all? :D And don't worry, I am working on a proper documentation ✌️ 

tl;dr its really uncool to wait for the HA release cycle to get a bug fixed that affects your pets

gjohansson-ST:
> you know that I am the developer of the official integration?^^ you know why this exists at all? :D And don't worry, I am working on a proper documentation ✌️
> 
> tl;dr its really uncool to wait for the HA release cycle to get a bug fixed that affects your pets

I did not find a PR in HA for this integration so that was my wondering.
Second it might not be bad as a custom rep, just wondering if it makes sense to have it in HACS default as it is in HA official.

benleb:
The "idea" to do it this way was born after the last HA update which broke the integration for many users. The integration interacts with some hardware devices and I do not own all of them and can therefore not really test the integration with them... After that event, the "beta" variant was introduced and is now used by some users I am in contact with via Github and the HA community forums. And of course, the plan is to "backport" new features and things to the official integration - but just after really being tested.

ludeeus:
As discussed in Discord, I'm closing this 👍

PR #978, opened 2021-06-07
Update integration

rsnodgrass:
Added Helium integration (brands and wheels-custom-integrations already updated)


rsnodgrass:
hassfest.yaml added

PR #980, opened 2021-06-10
Adds new integration [alexander0042/pirate-weather-ha]

alexander0042:
Submitting PirateWeather (https://pirateweather.net/) for inclusion in the default repository set. This integration is a copy of the Dark Sky sensor and weather module that does not rely on Dark Sky, providing a fully compatible alternative data source for any integration for users that do not have a Dark Sky API key or after the shutdown date.


ludeeus:
Can you please rename the repo?
It has nothing to do with hacs, thanks 👍

alexander0042:
Sorry about the naming issue, that makes complete sense now that you point it out! Corrected to pirate-weather-ha

ludeeus (changes requested):
CI does not pass

alexander0042:
Oh shoot, I though I’d fixed the [Python wheels issue](https://github.com/home-assistant/wheels-custom-integrations/pull/416) before I submitted this, but apparently not! I’ll take another look at that and hopefully solve it- sorry about this!

ludeeus:
see the issue...
You added `pirateweather` and not `darksky` (which you use as the domain)

`pirateweather` is a **much** better domain name for this integration, I suggest you rename that

alexander0042:
Thank you so much for clarifying that- I couldn't figure out why it wasn't passing, but that makes sense! I agree that a different domain sounds better, but part of my goal was trying to maintain comparability, so I was thinking of sticking with darksky. However, thinking about it more, I want to try to keep things as clear as possible. 

I'm going to change the name here, and then add a note to the docs that if people are trying to maintain 100% comparability, they'll need to change the name of the integration, or make a small tweak to whatever depends on it

ludeeus:
Sounds good! 👍

alexander0042:
I believe that this PR should be ready for review now! I've updated the brands and wheels to make sure that everything is passing on that end, as well as put a few rounds of improvements to the README to help with installation and setup

PR #1004, opened 2021-06-25
Add signalmessenger integration

MichaelBitard:


ludeeus (changes requested):
2 issues

- No information is given to the user.
- The repository contains other unrelated items (addons)

kevdliu:
Not sure if I did something wrong on my end but when I tried to install this via HACS it's being installed in the custom_components directory itself instead of its own subdirectory.

ludeeus:
The repo is not HACS compliant

PR #1014, opened 2021-07-04
Adds new integration [rroller/netgear]

rroller:
Add new integration [rroller/netgear]

https://github.com/rroller/netgear

netgear


rroller:
Brand is added in https://github.com/home-assistant/brands/pull/2668 
Waiting for that to be merged.

PR #1022, opened 2021-07-15
Adds new integration [madpilot/hass-amber-electric]

troykelly:
Removing my unmaintained one, replacing it with @madpilot 's hass-amber-electric


madpilot:
I'm the owner of the new integration - Let me know the best way to fix the above failure

PR #1028, opened 2021-07-18
Adds new theme [tgcowell/waves]

tgcowell:


theme
```diff
@@ -61,5 +61,6 @@
   "pbeckcom/green_slate_theme",
   "Poeschl/slate_red",
   "seangreen2/slate_theme",
-  "wowgamr/animated-weather-card"
+  "wowgamr/animated-weather-card",
+  "tgcowell/waves"
```
jbouwh, line 65: Please keep the list sorted

jbouwh (changes requested):
Please sort the list.

theme
tgcowell: Hi @jbouwh apologies I really should have noticed that before submitting! I have updated this as per your advice.

jbouwh (approved):

PR #1046, opened 2021-08-14
Add Upcoming anniversary

amaximus:
Add Upcoming anniversary


ludeeus (changes requested):
CI does not pass

PR #1048, opened 2021-08-16
Adds new integration [mullerdavid/hass_GreeExt]

mullerdavid:
Please add my custom component for extending gree climate components with swing mode service.
As it is using the builtin gree integration, wheel and brands tests were disabled in the automatic workflow checks.

### DIFF
--- a/integration
+++ b/integration
@@ -381,6 +381,7 @@
   "MTrab/landroid_cloud",
   "mudape/iphonedetect",
   "muhlba91/onyx-homeassistant-integration",
+  "mullerdavid/hass_GreeExt",
   "muxa/home-assistant-niwa-tides",
   "mvdwetering/huesyncbox",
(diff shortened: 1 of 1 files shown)


ludeeus (changes requested):
CI does not pass

mullerdavid:
> CI does not pass
> https://hacs.xyz/docs/publish/include/#before-submitting

As I indicated earlier, the wheel and brand tests are disabled. They are the failed ones. 

What should be the solution to resolve this? I can remove the dependency to skip the wheels test, as it is guaranteed to be installed, but I can't use the existing brand name/domain. Adding it to brand images again doesn't make sense.

ludeeus:
It's an independent integration, these things are needed

mullerdavid:
Added the icons and removed the requirement as the dependency has them. Should pass the checks now.

PR #1070, opened 2021-09-05
Adds new integration [jobvk/Home-Assistant-Windcentrale]

jobvk:
Add Windcentrale integration

Windcentrale provide an API to get realtime winturbine sensors and news messages.


ludeeus:
This seems to only be for users in a specific country?

PR #1076, opened 2021-09-12
Adds new integration [hif2k1/battery_sim]

hif2k1:


hif2k1:
I'm not sure why HACS Action is not successful. I thought I had done everything. Sorry if I'm wasting time, but I thought it was ok.

DurgNomis-drol:
@hif2k1 You need to remove this line from `hacs.json` -> `"content_in_root": true,` as content is not in root but situated as it should be 😃

hif2k1:
Many thanks I will amend.

PR #1098, opened 2021-10-08
Adds new integration [koolsb/ha-fpp]

koolsb:


ludeeus (changes requested):
CI does not pass.
https://hacs.xyz/docs/publish/include#before-submitting

PR #1112, opened 2021-10-26
Adds new integration [alryaz/hass-default-config-filter]

alryaz:
This is a filter for the `default_config` YAML option.

I have never seen anything like this so far, and probably for a good reason. However, `default_config` is a version-dynamic meta-component, that may (or may not) change between Home Assistant releases. An example given within README is the `energy` platform.

There are several edge cases not yet covered by this component (example: default_config not loading whenever a dependency is missing). These will be ironed out with future releases.

<hr>

P.S. While this integration provides means to configure it via a configuration flow, I am more than certain it does not deserve its own icon at `hass/brands`, therefore one of the checks will fail.


ludeeus (changes requested):
CI does not pass.
https://hacs.xyz/docs/publish/include#before-submitting

ludeeus:
This is overriding a core domain, that will not be accepted.
<https://hacs.xyz/docs/publish/include>

> Custom integrations that override core integrations will not be accepted, you can still use it as a custom repository

PR #1121, opened 2021-11-09
Adds new integration [dlashua/hass-wasp_sensor]

dlashua:


ludeeus (changes requested):
CI does not pass <https://hacs.xyz/docs/publish/include#before-submitting>

dlashua:
@ludeeus my apologies. I wrongly assumed ignoring "brands" in the repos actions would ignore them here too. I don't see much of a need for brands here, since the integration doesn't use config flow, but, I'll work on producing a logo and submitting a brands PR, then I'll open a PR here again.

PR #1129, opened 2021-11-16
Adds new integration [briis/hass-weatherflow]

briis:


briis:
This fails on `Requirements validation (more-info: https://hacs.xyz/docs/publish/include#check-requirements )`
I have read through the documentation many times now, and I simply can't see what I missing. Can you clarify?

PR #1137, opened 2021-11-25
Adds new integration [LAB02-Research/HASS.Agent-Notifier]

LAB02-Admin:
Add HASS.Agent-Notifier integration. GitHub Action checks passed.


LAB02-Admin:
Hi, is there a reason this isn't geting approved? Anything I can do/change?

PR #1150, opened 2021-12-18
Removed Integration briis/smartweather

briis:

### DIFF
--- a/blacklist
+++ b/blacklist
@@ -24,6 +24,7 @@
   "Bre77/myair",
   "brefra/home-assistant-plugwise-stick",
   "briis/mbweather",
+  "briis/smartweather",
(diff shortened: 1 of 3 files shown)


ludeeus (changes requested):
This changes more than it should

briis:
I followed the procedure [outlined here](https://hacs.xyz/docs/publish/remove) and the `make remove` command changed these files. So what is wrong?

Basically I just want to remove `briis/smartweather` as this has been replaced with a new Integration.

ludeeus:
That is the way, it looks like you have used an additional formatter on the `removed` file
You can try `make sort`

briis:
I will close this and start all over again.

PR #1161, opened 2021-12-25
Adds new integration [dvd-dev/hilo]

valleedelisle:
Submitting the integration for Hilo smart home product from Hydro-Quebec. We already have a good userbase, it would be easier for everyone if we can include this in HACS default.


valleedelisle:
Need to fork with user account as org accounts are not allowing edits by maintainers

PR #1179, opened 2022-01-01
Adds new integration [ardevd/ha-bobcatminer]

ardevd:


ardevd:
Brand icon has been added to the brand repository which should fix the action fail.

PR #1205, opened 2022-01-23
Remove bpost integration

sdebruyn:
Removes the bpost integration as bpost does not allow it


ludeeus:
This PR is missing 1 file, I have opened #1206 as a replacement to this PR

PR #627, opened 2020-09-21
Adds new integration [abacao/hass_wibeee]

abacao:
This PR adds the Wibeee component as a default component in HACS

PR #1227, opened 2022-02-01
Adds new integration [luuuis/hass_wibeee]

luuuis:
Originally added in https://github.com/hacs/default/pull/627, the abacao integration is not under active development any more whereas my fork has moved on significantly with improved reliability, multiple device support, energy support, a config flow for UI configuration and just keeping up with new HA versions in general ([see CHANGELOG](https://github.com/luuuis/hass_wibeee/blob/729f61b8bc3a4204851e17a36f1f60f10ba03cee/CHANGELOG.md)).

It uses semver, has tagged/zipped releases and I am planning further improvements such as Local Push support.


integration
```diff
@@ -4,7 +4,6 @@
   "9a4gl/hass-peltec",
   "9rpp/securifi",
   "AaronDavidSchneider/SonosAlarm",
-  "abacao/hass_wibeee",
```
ludeeus, line 7: This PR should not remove.
Removal needs to happen like this: <https://hacs.xyz/docs/publish/remove>

ludeeus (changes requested):

integration
luuuis: Sure thing, I will kick off a separate PR for the removal.

Is there a recommended way to handle transitioning users from one integration to another? I was thinking both integrations could coexist in HACS default for a while and in the meantime I can work with @abacao to maybe do a new release just to be able to inform users of the upcoming changes in the release notes. Unless there's a better way?

abacao: Let me know when there is a PR to remove my version

ludeeus: Remove when you want (assuming old still works)

luuuis: Removal PR is https://github.com/hacs/default/pull/1242

PR #1228, opened 2022-02-02
Adds new integration [thecode/ha-rpi_gpio]

thecode:

PR #1229, opened 2022-02-02
Add rpi_gpio integration

Sholofly:
As of version 2022.6.0 the core rpi_gpio integration will be removed from the core of Home Assistant. it's deprecated as of version 2022.2.0. This integration is en exact copy of the current core integration. Should we keep this one open until then or should I request an add then?

### DIFF
--- a/integration
+++ b/integration
@@ -648,5 +648,6 @@
   "youdroid/home-assistant-gitea",
   "youdroid/home-assistant-sickchill",
   "zachowj/hass-node-red",
-  "zigul/HomeAssistant-CEZdistribuce"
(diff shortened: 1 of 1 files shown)


ludeeus:
@thecode has already opened an "identical" PR (#1228) both these are failing CI, maybe you 2 should join forces?

Sholofly:
LOL, we'll do :)

thecode:
Opened as draft since waiting for the brands PR which conflicts with core

Sholofly:
@thecode You've placed your comment on my closed PR. I guess you would like to comment on your own?

PR #1238, opened 2022-02-08
Adds new integration [Raukze/home-assistant-fitx]

Raukze:


ludeeus (changes requested):
This service is only available in Germany and needs a country key in the hacs.json file.

Raukze:
Updated `hacs.json` file

PR #1247, opened 2022-02-13
Adds new theme [AmoebeLabs/HA-Theme_M3-04-Magenta]

AmoebeLabs:


AmoebeLabs:
Hmmm. I hope nothing goes wrong wrt the previous PR., as I see that one now in this PR.

I used a separate branch to make the PR against this branch, but my git-foo is not that strong 😄

PR #1258, opened 2022-02-20
Remove repositories with issues disabled

ludeeus:
These repositories no longer have issues enabled.

image
image
image


kukulich:
@ludeeus hmm, it looks I’ve disabled issues for https://github.com/kukulich/home-assistant-jablotron100 by accident. Should I send again PR to add the repository back?

ludeeus:
yes

EDIflyer:
@And3rsL would it be possible enable issues so the repo can be added back?

Friedrieck:
> @And3rsL would it be possible to enable issues so the Deebot repo can be added back into HACS?

FYI, there is a successor available in HACS (see https://deebot.readthedocs.io/integrations/home-assistant/migration/ for more info).
Take care of breaking changes

Friedrieck:
> Cant find this addon in HACS. What repo do I need to add?

It should be in the HACS repo itself (Deebot 4 Home Assistant). If it does not show up in the search results, maybe is it because it is already installed?

contactcr:
> > Cant find this addon in HACS. What repo do I need to add?
> 
> It should be in the HACS repo itself (Deebot 4 Home Assistant). If it does not show up in the search results, maybe is it because it is already installed?

You are right I was looking in Addons and not integrations, sorry.

PR #1262, opened 2022-02-22
Adds new integration [Bouni/abfallplus]

Bouni:


ludeeus (changes requested):
This integration is only available for users in Germany, and you should add a country key in hacs.json to set that

Bouni:
> This integration is only available for users in Germany, and you should add a country key in hacs.json to set that

Done!

PR #1265, opened 2022-02-24
Adds new integration [jnxxx/homeassistant-dabblerdk_powermeterreader]

jnxxx:
Adding an integration to read power consumption from Echelon/NES power meters through a Dabbler.dk MEP module.

### DIFF
--- a/integration
+++ b/integration
@@ -337,6 +337,7 @@
   "jihao/traccar-cn-hass",
   "jjlawren/sonos_cloud",
   "jm-73/Indego",
+  "jnxxx/homeassistant-dabblerdk_powermeterreader",
   "jobvk/Home-Assistant-Windcentrale",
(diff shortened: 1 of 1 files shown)


ludeeus (changes requested):
https://github.com/jnxxx/homeassistant-dabblerdk_powermeterreader/blob/main/hacs.json needs a country key

jnxxx:
Country has been added, but doesn't seem to be required according to the documentation: https://hacs.xyz/docs/publish/start#hacsjson

jnxxx:
It can be used in any country where these meters are being used, as long as they have a MEP port, so you can plug in the HW module developed by Dabbler.dk.
It is after all an American company producing these meters. So, likely they are also used elsewhere.

PR #1300, opened 2022-03-18
Adds new integration [iMicknl/ha-sagemcom-fast]

iMicknl:


iMicknl:
My bad, I totally missed that I never finished this... I will make sure that all checks are green and create a new PR when I have time.

PR #1309, opened 2022-04-03
Adds new integration [mbillow/ha-chargepoint]

mbillow:
Add ChargePoint custom integration to default repos.


ludeeus:
There are no instructions on how to use it.

mbillow:
> There are no instructions on how to use it.

Updating the README was actually on my todo list today. It really is just as simple and installing and logging in, though. 

Is there anywhere else that documentation needs to be?

mbillow:
I added installation, usage, and development instructions to the README. Let me know if there is anything else you want/need here.

PR #1319, opened 2022-04-11
Adds new integration [patrickribbing/sjofartsverket_viva-component]

patrickribbing:


patrickribbing:
That is correct, adding country to hacs.json.

PR #1593, opened 2022-11-27
Adds new integration [mtrab/clever]

MTrab:
https://github.com/MTrab/clever/actions/runs/3556239630/jobs/5986980619

https://github.com/MTrab/clever/actions/runs/3556239630/jobs/5973466307


integration
```diff
@@ -531,6 +531,7 @@
   "msekoranja/emsc-hacs-repository",
   "msp1974/homeassistant-jlrincontrol",
   "mtarjoianu/ha_lektrico",
+  "mtrab/clever",
```
ludeeus, line 534: <https://github.com/MTrab/clever/blob/master/hacs.json> Needs a country, this is for Denmark only.

ludeeus (changes requested):

integration
MTrab: Added

PR #1609, opened 2022-12-08
Adds new theme [th3jesta/ha-lcars]

th3jesta:
[**Link to successful HACS action:**  ](https://github.com/th3jesta/ha-lcars/actions/runs/3650851114)


th3jesta:
I believe that the last check failed because I generated my release before I had met those requirements. I regenerated it, so I'm guessing that check should pass, now.

PR #1613, opened 2022-12-11
"SPZB0001 Thermostat" no longer maintained, please remove

WolfRevo:
As I don't maintain "SPZB0001 Thermostat" any longer it should not be available in the official HACS store.

### DIFF
--- a/integration
+++ b/integration
@@ -787,7 +787,6 @@
   "wimb0/home-assistant-saj-modbus",
   "wizmo2/zidoo-player",
   "wlcrs/huawei_solar",
-  "WolfRevo/climate.spzb0001_thermostat",
   "xannor/ha_reolink_discovery",
   "xannor/ha_reolink_rest",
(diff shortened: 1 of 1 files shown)


integration
```diff
@@ -787,7 +787,6 @@
   "wimb0/home-assistant-saj-modbus",
   "wizmo2/zidoo-player",
   "wlcrs/huawei_solar",
-  "WolfRevo/climate.spzb0001_thermostat",
```
ludeeus, line 790: This is missing required changes <https://hacs.xyz/docs/publish/remove#remove-a-repository>

ludeeus (changes requested):

integration
ludeeus: Done in https://github.com/hacs/default/commit/8c8a5e792ce03d5c976aa6a5eeb69c2305e6e4af

PR #1625, opened 2022-12-20
Adds new integration [gndean/home-assistant-hypervolt-charger]

gndean:
Link to last HACS and Hassfest action: https://github.com/gndean/home-assistant-hypervolt-charger/actions/runs/3736500240/workflow


integration
```diff
@@ -318,6 +318,7 @@
   "gilsonmandalogo/hacs-minerstat",
   "gjohansson-ST/sector",
   "gjohansson-ST/stl",
+  "gndean/home-assistant-hypervolt-charger",
```
ludeeus, line 321: From the information I could find this is only for UK?
If so <https://github.com/gndean/home-assistant-hypervolt-charger/blob/main/hacs.json> needs country added

gndean: Yes, I think this is for UK only. I've added: `"country": "GB"` to hacs.json

Link to last HACS and Hassfest action: https://github.com/gndean/home-assistant-hypervolt-charger/actions/runs/3744571725/jobs/6358098196

PR #1631, opened 2022-12-27
Adds new integration [CM000n/qss]

CM000n:
https://github.com/CM000n/qss/actions/runs/3786242490/jobs/6436961896
https://github.com/CM000n/qss/actions/runs/3786242490/jobs/6436962005

### DIFF
--- a/integration
+++ b/integration
@@ -139,6 +139,7 @@
   "claudegel/sinope-gt125",
   "claytonjn/hass-circadian_lighting",
   "ClusterM/skykettle-ha",
+  "CM000n/QSS",
   "cnstudio/Taipower-Bimonthly-Energy-Cost-homeassistant",
   "CreasolTech/home-assistant-creasol-dombus",
   "crowbarz/ha-sql_json",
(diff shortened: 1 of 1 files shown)


CM000n:
Why do the "Run HACS Action" tests on this pull request fail with the following error message, while the HACS validation works within my repo?
`Error:  <Validation hacsjson> failed:  extra keys not allowed @ data['iot_class'] (More info: https://hacs.xyz/docs/publish/include#check-hacs-manifest )`

My `hacs.json` file also does not contain an entry with `iot_class` key, contrary to the error message.
I only use the iot_class key in my `manifest.json`.

integration
```diff
@@ -139,6 +139,7 @@
   "claudegel/sinope-gt125",
   "claytonjn/hass-circadian_lighting",
   "ClusterM/skykettle-ha",
+  "CM000n/qss",
```
ludeeus, line 142: https://github.com/CM000n/qss/blob/20221226-1/hacs.json
Needs a new release.

ludeeus (changes requested):

integration
CM000n: Thanks for the hint!

PR #1694, opened 2023-01-22
Remove stl integration

gjohansson-ST:
Removes `stl` integration as it's archived and not maintained.


ludeeus:
Removed in https://github.com/hacs/default/commit/d792e20abacceadf110bc4229d9279afdafd1068

PR #1699, opened 2023-01-23
Allow as Custom Repo [shbatm/hacs-isy994]

shbatm:
Remove shbatm/hacs-isy994 from the "removed" list. It is being used for a new beta test. Per guidelines this will not be added to HACS Default, but the removal message is giving people issues when installing it as a custom repo.


ludeeus:
<https://github.com/shbatm/hacs-isy994/blob/dev/.devcontainer/devcontainer.json#L2>
You should move away from that ([reason](https://github.com/ludeeus/container#deprecated)).
Example for an alternative:
- [`.devcontainer.json`](https://github.com/ludeeus/integration_blueprint/blob/main/.devcontainer.json)
- [scripts dir instead of `container` CLI](https://github.com/ludeeus/integration_blueprint/tree/main/scripts)

PR #2495, opened 2024-05-08
Remove archived repositories

ludeeus:


mtausk:
Regarding alexarch21/history-explorer-card... but why? Is it not working? I use it. Do you have any statistics how many people use it? Does it not work? Works for me. What is the reason for removal anyway? Is it just because of some policy?

ludeeus:
The reason is that it was archived by the author.

PR #3892, opened 2025-08-05
Adds new plugin [ArikShemesh/simple-timer-card]

ArikShemesh:
## Checklist

- [ ] (For integrations only) I've added the [hassfest action](https://developers.home-assistant.io/blog/2020/04/16/hassfest/) to my repository.

## Links

Link to current release: https://github.com/ArikShemesh/simple-timer-card/releases/tag/v1.1.4
Link to successful HACS action (without the `ignore` key): https://github.com/ArikShemesh/simple-timer-card/actions/runs/16742663732
Link to successful hassfest action (if integration): <>

### DIFF
--- a/plugin
+++ b/plugin
@@ -19,6 +19,7 @@
   "Anrolosia/Shopping-List-with-Grocy-Card",
   "arallsopp/hass-hue-icons",
   "argaar/comfortable-environment-card",
+  "ArikShemesh/simple-timer-card",
   "artem-sedykh/mini-climate-card",
   "artem-sedykh/mini-humidifier",
   "aukedejong/lovelace-windrose-card",


plugin
```diff
@@ -19,6 +19,7 @@
   "Anrolosia/Shopping-List-with-Grocy-Card",
   "arallsopp/hass-hue-icons",
   "argaar/comfortable-environment-card",
+  "ArikShemesh/simple-timer-card",
```
ludeeus, line 22: Why does this exist?
Would it not make sense to have it as a part of the integration as that is a requirement?

ArikShemesh: The tight coupling between card and integration is intentional. The card is specifically designed to expose all the unique features of the Simple Timer integration that generic cards cannot provide.
Additionally, some users experienced difficulties with manual installation (resource registration, file placement, etc.), so separating it into HACS allows for seamless installation and automatic resource management.

ludeeus: Im not talking about manual, your integration (like HACS) can automatically add this, I don't see why this needs to be a separate entry here, making it harder for the user by requiring them to do 2 things, instead of 1?

ArikShemesh: I didn't know this was a valid approach, mixing backend and frontend in HACS.
Could you point me to examples of integrations that include their frontend cards? I'd be happy to restructure this to provide a better user experience with single installation.

ludeeus: Ignore the HACS part of it. You want something for "custom integration." How that integration is downloaded is not relevant.

ArikShemesh: ok, will research.
thanks

ludeeus: There are several itnegraitons that setup icons for instance, those would work the same way.

ludeeus:
Converted to draft for now.

PR #4853, opened 2025-12-10
Adds new plugin [partach/switch_port_card_pro_card]

partach:
Adding card belonging to the switch-port-card-pro integration
https://github.com/partach/switch_port_card_pro_card

## Checklist

- [] (For integrations only) I've added the [hassfest action](https://developers.home-assistant.io/blog/2020/04/16/hassfest/) to my repository.

## Links

Link to current release: <https://github.com/partach/switch_port_card_pro_card/releases/tag/v0.8.0>
Link to successful HACS action (without the `ignore` key): <https://github.com/partach/switch_port_card_pro_card/actions/runs/20109654377>
Link to successful hassfest action (if integration): <>


plugin
```diff
@@ -349,6 +349,7 @@
   "OtisPresley/snmp-switch-manager-card",
   "ownbee/bootstrap-grid-card",
   "partach/switch_port_card",
+  "partach/switch_port_card_pro_card",
```
ludeeus, line 358: > This repository is only made because home assistant does not allow a integration-card combo installation.

That is not correct.
Integrations can register frontend elements.

HACS is an integration that does that, if it did not then adding it here, would also not work.

ludeeus (changes requested):

plugin
partach: Not sure if we talk about the same thing. 1st i made a lovelace card partach/switch_port_card (now old). This needed many yaml configuration additions that i did not like. Then i made an integration partach/switch_port_card_pro (this is the integration that does not need any yaml config... the name could be considered confusing in hind sight)
I have looked for a way to add the NEW! card i made to the integration but could not find a way that it installs correctly alongside the integration. Also asked the forum, no response on it.
What should be the structure of the integration code so that the NEW lovelace card will be installed and registered as well with the integration in 1 go?
Can you point me to the reference that explains this please?
I have seen a recent addition (last few weeks) that did exactly the same: have an integration and a card separate while the card is only meant for the integration... 
so there are now 3 artifacts:
- switch_port_card: still valid to use but needs much yaml work to get working and not flexible at all (only a dashboard)
- switch_port_card_pro: great new **integration** that handles all the snmp stuff for you and creates all needed entities on the fly. Could not find a way to register the added javascript file as card upon integration installation, so therefor;
- switch_port_card_pro_card: The new lovelace card alongside the integration as separate dashboard repository that can be hopefully installed via HACs (which was requested by users)

Any help is greatly appreciated as i am new to making these integrationx (and liking it very much)

Btw I tested this way and it works (via custom repository addition)

ludeeus: https://github.com/hacs/integration/blob/main/custom_components/hacs/repositories/plugin.py#L205-L230C1
That is how HACS itself does this (register a custom card from a custom integration).
Why can that not work for this?

As there is a hard link between the two elements, it does not make much sense to me that they should be separate entries needing downloading.

partach: Maybe we still are not talking about the same thing and please help me to understand.
Fully willing to do whatever is needed.

It seems other integrations to similar (separate card and integration repository). Is there a (very) recent change on this?
Or am i not understanding that in the integration i need to add a dist directory, place it there and let the user do 2 installs from the same repository (integration en then card?)

It seems technically possible from your end. But why is this nowhere mentioned in the 'standard' documentation?
(Maybe it is and I missed it, applogies. if it is avaible, please share)
Such document would help me understand much better what i need to do.
What i would need:
- in which directory structure inside the integration repository do i need to store the card (js)?
- will this then automatically be copied inside /www/community/switch_port_card_pro/here?
- Or is it installed somewhere else in this case? in the custom_components directory together with the integration?
- Will it register this resource in Dashboard? (so when i declare type: custom:switch-port-card-pro it will recognize it). Reason to ask 'normally' cards are stored in /local/communities/...

Any help really appreciated, can we DM on this in some way?

ludeeus: There is nothing standard about custom integrations, so there is little to no documentation anywere.
` /www/community/` This is what HACS uses; your integration can use that as well, or even better, something completely different.
If it is under /www/ HA, set up the public listing for it, so you only need to copy it there when your integration starts and register it.
or you just keep it in your integration dir and register it from there with some other handler.

This has always been the case. I have seen some posts that state this is not possible, which has resulted in unnecessarily many repositories and frustration for the user, as they need 2 separate things to have a single thing work properly.

partach: I have it and had it in the /www/community directory in my integration repository. Yet at installation testing in my environment the standard install did not copy it automatically? Why wouldnt HA automatically copy it to the target in /www/community if it sees that directory in the integration directory structure and register it? My own preference is that i have one repository with card and integration as that was always the goal. Is there an example of the custom community that made this work i could look at? (their github). As we say in Dutch, a good example makes followers :)
IS there a code example that copies this toward the desired directory and code example to register it myself? If there is i am happy to implement it

partach: not knowing the details yet

ludeeus: There is no automation.
You need to manually copy it.
How you have it in your repository does not matter, and HA/HACS does not know of anything other than your custom integration file.

This all needs to be added as code by you.

ludeeus: HACS itself register and serves its own front-end for instance.
Most of the custom icon integrations works the same way.

partach: > There is no automation. You need to manually copy it. How you have it in your repository does not matter, and HA/HACS does not know of anything other than your custom integration file.
> 
> This all needs to be added as code by you.

I don't mind to code myself but there is no real supported documentation i could find that explains? Trying to figure out by guessing based on searching examples which is not great..
Btw if you place the js file in the custom_componentens directory of the repository (as per examples i found like with icons), i don't see why it isnt copied in your local HA... it will?

I placed the js file in my repository at /custom_components/frontend/.
I will leave it in this directory on the local HA, no copy needed.
So only thing left to do: find which api calls are needed to register it correctly....

Am i close?
```
CARD_URL = f"/{DOMAIN}/switch-port-card-pro.js"
CARD_JS = f"custom_components/{DOMAIN}/frontend/switch-port-card-pro.js"

async def async_setup(hass: HomeAssistant, config: dict) -> bool:
    hass.data.setdefault(DOMAIN, {})

    try:
        await hass.http.async_register_static_paths([
            StaticPathConfig(CARD_URL, hass.config.path(CARD_JS))
        ])
        _LOGGER.info("Switch Port Card Pro card served at %s", CARD_JS)

        add_extra_js_url(hass, CARD_URL)

        async def _register_resource(event=None):
            try:
                from homeassistant.components.lovelace import resources
                
                # Get current resources
                resource_collection = await resources.async_get_resource_collection(hass)
                
                # Check if already registered
                existing = [r for r in resource_collection.async_items() if r.get("url") == CARD_JS]
                
                if not existing:
                    # Add the resource
                    await resource_collection.async_create_item({
                        "res_type": "module",
                        "url": CARD_JS,
                    })
                    _LOGGER.info("Auto-registered Switch Port Card Pro as Lovelace resource")
                else:
                    _LOGGER.debug("Switch Port Card Pro already registered")
                    
            except Exception as err:
                _LOGGER.warning(
                    "Could not auto-register Lovelace resource. "
                    "Please add manually: %s", 
                    err
                )
    
        hass.bus.async_listen_once("lovelace_updated", _register_resource)

    except Exception as err:
        _LOGGER.warning("Frontend registration failed for %s", CARD_JS)
    return True

```

I must be getting close as it seems i am almost there... 
If i now do it like above and then register the card in the UI with 
Settings → Dashboards → 3 dot menu → Resources → Add resource:
URL:
custom_components/switch_port_card_pro/frontend/switch-port-card-pro.js
Resource Type:
Java script.
It works! ...On my machine and one of the users reported this is working!
How do i do this last step via the code above?

partach: Hope you can help with direction on doing the resource registration via code!

partach: Just had a small chat with Frenck...  He says it is not possible.... Please prove him wrong ;)

partach: hhmm on mobile platform it doesnt work great. In browser it needs an extra refresh to show but in companion app you can't do that. So it seems conclusion is coming to there is no good way other then separate card and integration :(

partach:
Solved it myself

PR #5399, opened 2026-02-05
Add diestrohs/time-spinner-card

diestrohs:
Add the Time Spinner Card to the HACS plugin default list.


ludeeus:
This is all sorts of invalid, and the PR template was not used.

PR #6521, opened 2026-03-25
Adds new integration [TheBullRing/home-assistant-glucose-ng]

TheBullRing:
Adding integration TheBullRing/home-assistant-glucose-ng to default integrations

## Links

Link to current release: <https://github.com/TheBullRing/home-assistant-glucose-ng/releases/tag/v1.0.1>
Link to successful HACS action (without the `ignore` key): <https://github.com/TheBullRing/home-assistant-glucose-ng/actions/runs/23512253812>
Link to successful hassfest action (if integration): <https://github.com/TheBullRing/home-assistant-glucose-ng/actions/runs/23512253814>


frenck (changes requested):
Hi @TheBullRing, thanks for the submission! 👋

A couple of items to address before this can be added to the HACS default catalog. The token-validation pattern on your POST views is fine as a legitimate webhook-style carve-out from HA's session auth, so those stay in scope as designed. Two things to fix:

**1. `__pycache__/` committed to git.** The directory `custom_components/glucose_ng/__pycache__/` contains compiled Python bytecode that should not be in the repository. HACS ships everything under `custom_components/<domain>/` to users, so those `.pyc` files reach every install. Please add `__pycache__/` and `*.pyc` to `.gitignore`, remove the directory from the tracked tree, and cut a new release.

**2. HTTP routes registered under HA's reserved `/api/` namespace.** In `http.py` you register nine views at paths like `/api/v1/entries`, `/api/v3/entries`, `/api/v1/treatments`, `/api/v1/status`, `/api/v2/authorization/request/{token}`, `/api/v3/version`, etc. The `/api/*` URL space belongs to Home Assistant's own REST API. Today none of those specific Nightscout paths collide with a core HA endpoint, but the namespace is reserved and a future HA release could collide with one of them; integrations should not register routes inside `/api/`.

The right pattern is to expose the Nightscout-compatible surface under a non-reserved prefix that the user can point uploaders at. For example, mount everything under `/glucose_ng/v1/entries`, `/glucose_ng/v3/entries`, `/glucose_ng/v1/status`, etc. Uploaders like Juggluco and xDrip accept a configurable base URL, so the user enters `https://<ha-host>/glucose_ng` as the Nightscout server URL and the integration's relative paths line up.

If you want the URL prefix to be user-configurable (e.g., `/nightscout` for users who prefer that), expose it as a config-entry option and parameterize the view registration accordingly.

Once those are addressed, please cut a new release and re-request review.

../Frenck
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

TheBullRing:
All requested changes have been implemented.

A new release has been published: v1.0.2

Ready for re-review ✅

TheBullRing:
@frenck changes were made a month ago, am I missing anything to get this published?

frenck (changes requested):
Hi @TheBullRing, thanks for clearing out `__pycache__` and moving the routes behind a configurable prefix.

Two things still need work, and the first one I got wrong in my last review, so let me correct myself: I described the token check on your POST views as a legitimate webhook style carve-out. Looking at the current code more carefully, the validation is not doing what it appears to.

1. **The authentication can be bypassed with a request header.** `_check_auth` grants a five minute session keyed on the client IP, and the IP is taken from headers the client controls, in `http.py` lines 40 to 46:

```python
def _get_client_ip(request: web.Request) -> str:
    return (
        request.headers.get("X-Real-IP")
        or request.headers.get("X-Forwarded-For", "").split(",")[0].strip()
        or request.remote
        or "unknown"
    )
```

Anyone who can reach the endpoint and set `X-Real-IP` to an address that authenticated in the last five minutes is let straight through without presenting the shared secret. Alongside that, `GlucoseNGStatusView` (line 561) and `GlucoseNGVersionView` (line 664) call no authentication at all, and `GlucoseNGV2AuthView` returns HTTP 200 with the submitted token echoed back even when it matches nothing (line 543), which also logs that token in cleartext at warning level, on by default. Your README asks users to expose Home Assistant to the internet, and the payload here is continuous glucose data, so this needs to be a real check: drop the IP session entirely and require the shared secret on every request, or move to Home Assistant's `webhook` component so the per-webhook token is validated by core.

2. **The reserved namespace is still reachable.** The configurable prefix is the right fix, but an empty prefix is still accepted and mounts everything back under bare `/api/v1/...`, per the comment and code at lines 684 to 708. Please require a non-empty prefix so the `/api/` space is never used.

Separately and non-blocking: `manifest.json` declares version `0.2.0` at tag `v1.0.2`, and `DeviceInfo` hardcodes `sw_version="0.3.0"` in both `sensor.py` and `event.py`, so there are three different version strings in one release.

Once those are addressed, cut a new release and re-request review.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

TheBullRing:
Thank you for the detailed review. All the issues you raised have been addressed in release v1.0.3. Here's what was done:

**Authentication bypass via IP session — fixed**
The entire IP-session mechanism (_auth_sessions, _grant_session, _check_session, _get_client_ip) has been removed. Every request must now present the shared secret directly on each call. There is no cached state that can be spoofed via X-Real-IP or X-Forwarded-For.

**GlucoseNGV2AuthView returning 200 on any token — fixed**
The endpoint now returns HTTP 401 when the submitted token does not match any registered entry. The token is no longer echoed back in the warning log.

**GlucoseNGStatusView and GlucoseNGVersionView unauthenticated — fixed**
Both endpoints now call _check_auth on every request and return 401 if the shared secret is not presented.

**Empty prefix allowing /api/ namespace collision — fixed**
An empty url_prefix is now rejected both in the config flow (inline validation error) and in register_http_views (raises ValueError). The /api/ namespace can never be used.

**Version string inconsistency — fixed**
manifest.json, sensor.py, and event.py all declare version 1.0.3.

**Could you please re-review when you have a moment? Thank you.**

PR #7289, opened 2026-04-27
Rename repositories

ludeeus:


ludeeus:
CI failure is expected, I will fix that later.

JohnLindahlTech:
Oh, so this happens automatically. I created https://github.com/hacs/default/pull/7280 because I thought it was needed, but this pr will make that happen automatically then?

ludeeus:
It is not automatic, but handled for you https://hacs.xyz/docs/publish/update/

PR #8181, opened 2026-06-02
Adds new integration [mosandlt/Bosch-Smart-Home-Camera-Tool-HomeAssistant]

mosandlt:
## Checklist

## Links

Link to current release: https://github.com/mosandlt/Bosch-Smart-Home-Camera-Tool-HomeAssistant/releases/tag/v13.5.3
Link to successful HACS action (without the `ignore` key): https://github.com/mosandlt/Bosch-Smart-Home-Camera-Tool-HomeAssistant/actions/runs/26843459915/job/79156563376
Link to successful hassfest action (if integration): https://github.com/mosandlt/Bosch-Smart-Home-Camera-Tool-HomeAssistant/actions/runs/26843459915/job/79156563395


frenck (changes requested):
Hi @mosandlt, thanks for the submission! 👋

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

frenck (changes requested):
Thanks for submitting this, and for the amount of work that clearly went into the Bosch Smart Home camera side of things. The tunnel and FCM plumbing is not trivial, and it shows.

I went through the code at the release tag and there are three things I need addressed before this can go into the default catalog.

### 1. TLS verification is disabled toward a cloud endpoint

`custom_components/.../tls_proxy_wiring.py` lines 44 to 46:

```python
ctx.check_hostname = False
ctx.verify_mode = ssl.CERT_NONE
```

That context is what `live_connection.py` lines 427 to 451 uses to reach `proxy-NN.live.cbs.boschsecurity.com`. That is not a device on the local network with a self signed certificate, it is a fixed hostname on the public internet, and the payload is live video from someone's home. With hostname checking and verification both off, anything able to intercept that connection can present any certificate and read or modify the stream.

For local Bosch controllers with their own self signed certificate, pinning the expected certificate (or the controller's CA) is the right answer. For the cloud proxy hostnames, please use normal verification against the system trust store.

### 2. Snapshots are written into `config/www`

`fcm.py` line 2734 writes alert snapshots into `config/www/bosch_alerts/`. Everything under `config/www` is served at `/local/` with no authentication at all, so any snapshot the integration saves becomes readable by anyone who can reach the Home Assistant HTTP port, including unauthenticated. For camera images from inside a house that is a meaningful exposure, and the filenames are guessable.

Please write these somewhere outside `www`. If the images need to be surfaced in the frontend, an image or camera entity, or a view that checks the request's user, gives you the same result without the public directory.

### 3. No admin gate on the services

None of the registered services check whether the calling user is an administrator. That includes `BoschPanicAlarmSwitch` and the arm and disarm paths. `hass.services.async_register` handlers are reachable by every authenticated user, including non admin users, so as it stands a non admin household member can disarm the alarm or trigger the panic alarm.

For anything that changes the state of an alarm system, please check the calling user:

```python
user = request.get("hass_user")
if user is None or not user.is_admin:
    raise Unauthorized()
```

or, for service calls, resolve the context user through `hass.auth` and reject non admins. Note that a check written as `if user is not None and not user.is_admin` lets an unresolved user straight through, so the `None` case has to fail closed.

Happy to take another look once those three are sorted.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

mosandlt:
Hi @frenck, thanks for taking the time on this — really appreciate the depth of the review. Fixed all three, released as [v16.1.14](https://github.com/mosandlt/Bosch-Smart-Home-Camera-Tool-HomeAssistant/releases/tag/v16.1.14), commit [`a469468`](https://github.com/mosandlt/Bosch-Smart-Home-Camera-Tool-HomeAssistant/commit/a4694688a02001fcafa8a2dd40651b1446509bf2).

**1. TLS verification on the cloud proxy**

You're right that `CERT_NONE`/`check_hostname=False` on a public-internet host was wrong. Fixed it to require a real certificate chain — but a plain system-trust-store `CERT_REQUIRED` isn't enough here: Bosch's cloud/video-proxy hosts (`*.cbs.boschsecurity.com`) turn out to be served by a *private* Bosch PKI, not a publicly trusted CA (we already pin that same CA for the REST API in `cloud_ssl.py`, just hadn't reused it for the TLS proxy). So the proxy now builds its context from that pinned Bosch CA (`Bosch ST Root CA` → `Video CA 2A`) with `VERIFY_X509_PARTIAL_CHAIN`, `CERT_REQUIRED`. `check_hostname` does stay off — the `proxy-NN.live.*` hostname genuinely doesn't match the cert's SAN (`*.residential.connect.boschsecurity.com`), which is also why go2rtc's own Go TLS client rejects it when hostname checking is attempted — but the cert now has to chain to Bosch's actual CA, so an attacker can't just present any certificate anymore.

Small aside since it's relevant to trusting this: I initially shipped this with a bare system-trust-store context (matching your literal suggestion) and it would have quietly broken every REMOTE stream in production — caught before release by three independent reviewers, one of whom did a live TLS handshake against `residential.cbs.boschsecurity.com` to confirm the pinned-CA version works and the bare version doesn't. Wanted to flag that in case it's useful context for how literally to read "use normal verification" suggestions against services with private PKI.

**2. Snapshots served from `config/www`**

Kept them there, for one specific reason: HA's mobile_app push notification image is fetched by the OS's own push handler, outside any Home Assistant session — there's no auth header it can carry, so it genuinely can't sit behind an authenticated view (this is the same reason most integrations that attach an image to a push notification use `/local/`). What I could and did fix is the actual guessability: filenames were `{camera}_{timestamp}_{event}.jpg`, which is enumerable. They now carry a 128-bit random token per alert (`secrets.token_urlsafe(16)`) on top of that, so there's nothing left to guess even though the file is technically public and HA doesn't expose directory listing on that route. Files are still deleted ~30s after delivery by default; if the user opts into "keep for reference" they stay named the same random-token way.

If you still see this as insufficient for the default catalog I'm open to it — happy to look at moving to a signed-path view as a next step, just wanted to explain the constraint rather than silently deviate from your suggestion.

**3. No admin gate on the panic alarm / arm-disarm**

Added an admin check on both — `BoschPanicAlarmSwitch` (physical 75dB siren trigger) and the intrusion-system arm/disarm switch now resolve the calling user via `self._context.user_id` → `hass.auth.async_get_user()` and raise `Unauthorized` if that user isn't an admin (or doesn't resolve at all — fails closed, not open). Calls with no user context — automations/scripts — pass through, since only admins can author those in HA in the first place, so that's already an admin-gated surface one level up.

All three have new regression tests, mypy --strict / ruff / codespell clean, 6834 tests / 100% coverage. Let me know if any of this needs another look.

frenck (changes requested):
Hi @mosandlt, thanks for turning v16.1.14 around so quickly, and for the detailed write-up on the reasoning behind each change! 👋

The pinned-CA work on the cloud proxy and the token-based snapshot filenames both answer what I raised, so I'm not going to hold those. Two things are still open:

1. The admin gate does not do what it needs to. `switch.py:169-175` reads `user_id = context.user_id if context is not None else None` and then `if user_id is None: return`, so a call with no resolvable user passes straight through. That is the case I asked to fail closed, because it is the one an unauthenticated or context-less caller produces. Separately, `is_admin` appears only in `switch.py` in the whole integration, so the gate covers exactly two entities, arm/disarm and panic. The roughly twenty-five services registered at `services.py:1756-1828` have no admin check at all, and several of them are destructive or account-level: `share_camera`, `invite_friend`, `remove_friend`, `delete_event`, and the `create_rule`/`update_rule`/`delete_rule` set. Please make the gate reject a missing user and apply it to those services too, either through `async_register_admin_service` or the same explicit check.

2. `__init__.py:983-985` calls `cf_unbuffer.register(hass)` unconditionally at setup, and `cf_unbuffer.py:212-224` rebinds `handle` on HA core's `HlsMasterPlaylistView`, `HlsPlaylistView`, `HlsInitView`, `HlsPartView` and `HlsSegmentView` class objects. Because those are class attributes, that changes HLS Content-Type and transfer encoding for every camera stream in the instance, not just Bosch ones, for as long as the process lives, and there is no way for a user to turn it off. Please put it behind an explicit option that defaults to off, and say in the option's description what it changes.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

mosandlt:
Thanks for the follow-up — both fair, and you were right that round 1 didn't actually close the gap you asked for. Fixed, released as [v16.1.15](https://github.com/mosandlt/Bosch-Smart-Home-Camera-Tool-HomeAssistant/releases/tag/v16.1.15), commit [`486d978`](https://github.com/mosandlt/Bosch-Smart-Home-Camera-Tool-HomeAssistant/commit/486d978f7c17bb9a3dce719f716c6867001798bf).

**1. Admin gate**

You're right, the missing-user_id case was exactly backwards — I'd modeled it on HA core's own `helpers.service.async_register_admin_service`, which does treat a missing `context.user_id` as an implicitly-trusted automation call, but that's not good enough for a physical siren / intrusion-arm switch. It now rejects unconditionally: no context, no user_id, an unresolvable user_id, or a non-admin user are all `Unauthorized`.

Also extended it to the ~20 custom services at `services.py:1756-1830`, with one deliberate exception: `trigger_snapshot`. Before shipping I ran an adversarial bug-hunt on this exact diff, and it caught that the Lovelace card calls `trigger_snapshot` automatically on page load, tab switch, and a 60s timer for every viewer — admin-gating it would have broken live-image refresh for any non-admin household member using the dashboard. It's a read-only refresh trigger, not account-level, so I left it out and documented why in the code. Flagging this since it's a case where I didn't apply your suggestion literally across the board — happy to hear if you see it differently.

**2. cf_unbuffer**

Now behind a new option (`cloudflare_tunnel_hls_unbuffer`, default off, config-flow description states the instance-wide scope explicitly, matching what you asked for).

One thing the same bug-hunt caught that's worth flagging: `cf_unbuffer`'s wrapped view classes have a second job I hadn't separated out — they're the only source for a small access-timestamp cache (`hls_access_age`) that the stream-lifecycle idle-reaper reads to detect "is someone actively watching this camera over HLS right now." My first pass at this fix gated the entire `register()` call behind the new option, which would have silently killed that detection for every default (option-off) install — the reaper would think nobody's watching and tear a session down mid-viewing. Fixed by splitting it: the wrapper *installation* (and the access-timestamp stamping) stays unconditional; only the actual Content-Type/Transfer-Encoding rewrite is gated behind the option now.

6843 tests / 100% coverage / mypy --strict / ruff / codespell clean. Let me know if either of these still needs another pass.

frenck (approved):
Hi @mosandlt, thanks for taking both of those seriously, and for the reasoning you left in the code alongside the fixes! 👋

Approving and merging now.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

PR #8522, opened 2026-06-15
Skip repository-specific checks for renamed-repositories PRs

ludeeus:
- Guard the preflight job (and via the needs cascade, all repository-specific jobs) with a label condition so that PRs labelled 'renamed-repositories' skip the new-addition validation.
- Add the 'labeled' trigger type so applying the label re-evaluates the workflow.


ludeeus:
Not needed with #8524

PR #8709, opened 2026-06-23
Adds new integration [pacorola/Atmoce_battery_HA]

pacorola:
## Checklist

## Links

Link to current release: <https://github.com/pacorola/Atmoce_battery_HA/releases/tag/v0.1.2>
Link to successful HACS action (without the `ignore` key): <https://github.com/pacorola/Atmoce_battery_HA/actions/runs/24859904574>
Link to successful hassfest action (if integration): <https://github.com/pacorola/Atmoce_battery_HA/actions/runs/24859904594>


pacorola:
Good morning @frenck , I had to redo the pull request since when I requested review on the old one it closed automatically with unmerge. Thanks for your comment! I fixzed the entry per your request.

frenck (changes requested):
Hi @pacorola, thanks for the submission! 👋

`diagnostics.py` returns the config entry data without redacting `CONF_CLOUD_WEB_PASSWORD`, so the user's cloud account password ends up in plaintext in the downloadable diagnostics file. Since diagnostics are frequently attached to public GitHub issues, this exposes the password. Please add `CONF_CLOUD_WEB_PASSWORD` (and any other secrets) to the redaction set, for example via `async_redact_data(entry.data, TO_REDACT)` with the password key included in `TO_REDACT`.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

pacorola:
Fixed in v1.3.0! Hope it now checks all boxes, thanks Frenck!

frenck (changes requested):
Hi @pacorola, the diagnostics ask is properly fixed — `TO_REDACT` now covers `CONF_CLOUD_WEB_PASSWORD` alongside the app key, app secret and email, and both `entry.data` and `entry.options` go through `async_redact_data`. Thanks. 👋

Two things left.

1. The cloud API secret is typed in clear while the password next to it is not. In `config_flow.py`, `STEP_FALLBACK_SCHEMA` and `AtmoceOptionsFlow.async_step_init` both declare:

```python
vol.Optional(CONF_CLOUD_APP_KEY, default=""): str,
vol.Optional(CONF_CLOUD_APP_SECRET, default=""): str,
```

You already have `_PASSWORD_SELECTOR` and apply it correctly to the web password, so this looks like an oversight rather than a decision — `CONF_CLOUD_APP_SECRET` should use the same selector.

2. No Home Assistant floor is declared anywhere. Neither `hacs.json` (which is just `{"name": "Atmoce Battery", "render_readme": true}`) nor the manifest sets one, while `config_flow.py` uses `ConfigFlowResult` (2024.4), calls `self._get_reconfigure_entry()` and constructs `AtmoceOptionsFlow()` with no entry, relying on the implicit `OptionsFlow.config_entry` from 2024.11. Please add `"homeassistant": "2024.11.0"` to `hacs.json` so HACS can keep it off installs that would break.

Smaller, for the same release: `manifest.json` has moved on to `1.5.1`/`1.5.2` while the newest release is tagged `v1.5.0`, so the version users install does not match what the manifest claims. And the rolling `nightly` prerelease that re-publishes on every commit to main is fine for your own testing, but worth being aware that a user who flips on prereleases in HACS will be tracking main.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

pacorola:
Thanks for your keen eye Frenck!!!! All fixed and implemented your best practices to nightly! (now only releases if i make unpublished changes) and better fits my development cycle. Awesome suggestions!

frenck (changes requested):
Hi @pacorola, thanks for masking the app secret, declaring the 2024.11.0 floor and getting the manifest version back in step with the tag! 👋

Two things the full pass turned up.

1. The declared `pymodbus` floor is four minor versions below the API the code actually calls. `manifest.json` says `"pymodbus>=3.6.0"`, but `modbus_client.py` passes `device_id=` on every read and write (lines 108-110, 143, 152, 161). That keyword only arrived in pymodbus 3.10.0. I unpacked the wheels to check rather than going from memory: on 3.6.x the signature is `read_holding_registers(self, address, count=1, slave=0, **kwargs)`, so `device_id=` is silently swallowed into `**kwargs`, never reaches `slave`, and every read and write addresses unit 0 instead of the configured slave. On 3.7 through 3.9 there is no `**kwargs` at all, so it raises `TypeError`, and none of your handlers catch that, so setup dies with "Unknown error occurred". This is not a comment on `>=` versus `==`; the operator is fine, the version number is just too low for the argument you use. It matters in practice because pip leaves an already-satisfying older pymodbus in place, and Home Assistant's own `modbus` integration is a common source of one.

2. Raw portal payloads reach the debug log while diagnostics deliberately redacts the same object. `web_client.py` line 115 logs the entire `selectModel` response and line 133 logs the full `changeModel` request and response. That is the same `storageModel` that `diagnostics.py` refuses to emit unredacted: `_MODEL_REDACT_HINTS` at lines 32-46 strips `mail`, `address`, `owner`, `name`, `latitude`, `longitude`, `token`, `user` and `account`, with your own comment at lines 28-31 explaining that an address reaching a public issue is not a nuisance. Debug logs get attached to public issues at least as often as diagnostics downloads do. Running those two lines through `_redact_model`, or just dropping the payload from the message, closes it. Same file, line 75 logs the cloud account email that `TO_REDACT` classifies as a secret.

Two non-blocking notes for the same release. `custom_components/atmoce/icon.png` and `custom_components/atmoce/brand/icon.png` are two *different* 256x256 images shipping in the same package, so which one a given core renders decides which icon users see; `brand/` is the current path and the root copy looks like the leftover. And `DEFAULT_SCAN_INTERVAL = 10` is applied unconditionally, including on the cloud fallback path at `coordinator.py` lines 218-222, so while the Modbus gateway is down the integration hits Atmoce's Open API every ten seconds for data your own comment describes as having fifteen-minute latency.

On the nightly prerelease: I see you already fixed the churn on main in `177b636`, after the tag. It is flagged prerelease so default-channel users never see it, and it was never a blocker.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

pacorola:
Thanks again @frenck !!!! Real keen eye you have!

frenck (changes requested):
Hi @pacorola, thanks for raising the pymodbus minimum to 3.10.0, adding the shared `redact_model` helper so the debug log and diagnostics now follow one rule, dropping the stray package-root `icon.png`, and throttling the cloud fallback to its own 15 minute interval! 👋

One thing left, and it is a version declaration rather than code.

`hacs.json` declares `"homeassistant": "2024.11.0"`, one release short of what the options flow needs. `custom_components/atmoce/config_flow.py` line 373 returns `AtmoceOptionsFlow()` with no argument, and the class at line 376 never assigns `self.config_entry`, so it relies entirely on the implicit `OptionsFlow.config_entry` property when it reads the entry at lines 402, 403 and 409. Home Assistant only populates that property from 2024.12. On a 2024.11.x core, HACS still offers and installs the integration because the declared floor says it is compatible, and setup and polling work normally, so nothing looks wrong until the owner presses Configure to enter or change the atmocecloud.com login or the Open API credentials. The options step then raises before the form is built and the dialog dies with "Unknown error occurred", leaving no way to add the portal login and therefore no battery SOC limit entities, with nothing pointing at the core version as the cause.

The code itself is the right shape: the argument-free constructor plus the implicit property is exactly what works on current cores, since that setter was removed in 2025.12. So the fix is `"homeassistant": "2024.12.0"` in `hacs.json`, not a change to `config_flow.py`.

Two non-blocking notes for the same release. `redact.py` line 26 lists `"lat"` as a bare substring hint, so it also redacts unrelated keys such as `latestVersion`; harmless over-redaction, but a more specific hint would keep diagnostics readable. And `coordinator.py` lines 116 to 122 call `super().__init__` without `config_entry=` and then assign `self.config_entry` afterwards; that still works, but recent cores want the entry passed into the constructor and this is the shape that tends to start emitting deprecation warnings for custom integrations.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

frenck (approved):
Hi @pacorola, thanks for raising the floor to 2024.12.0! 👋

Approving and merging now.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

pacorola:
Awesome! Proud and happy to be part of this great community and thanks a lot for your support!!!

Enviado desde Outlook para Android<https://aka.ms/AAb9ysg>
________________________________

PR #8891, opened 2026-07-02
Adds new integration [Lex-campbell/centsys_remote]

Lex-campbell:
## Checklist

## Links

Link to current release: <https://github.com/Lex-campbell/centsys_remote/releases/tag/v0.3.0>
Link to successful HACS action (without the `ignore` key): <https://github.com/Lex-campbell/centsys_remote/actions/runs/28597337676/job/84796104160>
Link to successful hassfest action (if integration): <https://github.com/Lex-campbell/centsys_remote/actions/runs/28597337676/job/84796104437>

Made with [Cursor](https://cursor.com)


frenck (changes requested):
Hi @Lex-campbell, thanks for the submission! 👋

In `api/mqtt_remote.py` the MQTT TLS context is built with `ssl.CERT_NONE` and `check_hostname = False`, so the server certificate of the cloud broker is never verified. Because this channel opens and closes a physical gate, a man-in-the-middle on the path to the broker could observe or tamper with those commands. The client-side mTLS authenticates you to the broker, but it does not protect against an unverified/forged server.

Please verify the broker certificate rather than disabling it, for example by loading the broker's CA into the SSL context (`load_verify_locations`) and leaving `check_hostname`/`CERT_REQUIRED` on. If the broker uses a private CA, ship/pin that CA rather than turning verification off.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

Lex-campbell:
Thanks @frenck — addressed in [v0.4.5](https://github.com/Lex-campbell/centsys_remote/releases/tag/v0.4.5).

MQTT TLS now pins Centurion’s private CA (`api/certs/centsys_ca.pem`) and verifies the broker leaf (`CN=CentsysQA`) with `CERT_REQUIRED` / hostname checking on. `CERT_NONE` is gone.

Ready for re-review.

frenck (changes requested):
Hi @Lex-campbell, the TLS fix is exactly right, and stricter than what I asked for: pinning the private CA through `create_default_context(cadata=...)` narrows the trust anchor to that one certificate, `CERT_REQUIRED` and hostname checking are both left on, and you applied it to all three MQTT paths rather than just the one I quoted. Relaxing only `VERIFY_X509_STRICT` for the missing Authority Key Identifier, with the reason in a comment, is the right narrow exception. 👋

The full pass turned up two things, and given that this integration opens driveway gates and garage doors I want to be careful about both.

1. `custom_components/centsys_remote/api/const.py` lines 24-31 ship a hardcoded production credential. That JWT is Centurion's live service-level bearer for the production gateway, it does not expire until 2055, and its payload carries what looks like a shared secret in the `hash` claim. `api/client.py` line 290 uses it as the last-resort bearer for `SendOtp` and `ValidateOtp`, which is to say it is the thing that lets an unauthenticated caller make Centurion send a one-time code to any phone number registered on their system. Shipping it in a catalog integration republishes it to everyone who installs.

   I would rather understand the constraint than just tell you to delete it, because I can see removing it breaks OTP login outright and there may be no user-supplied alternative. So: is there a sanctioned way to obtain a bearer for those two calls, and have you been able to raise this with Centurion? If the honest answer is that the integration only works by carrying a credential lifted from their app, that is a conversation worth having with them before this goes into the default catalog, and I am happy to leave the PR open while it happens.

2. Credentials reach the log. `api/client.py` lines 224-227 log the request URL and body, and lines 247-250 log the full response body, with only the `Authorization` header redacted. Every credential-bearing call goes through that one helper, so in practice: `validate_otp` logs the submitted OTP on the way out and the long-lived session JWT on the way back, since the response body *is* the token; `get_certificate` logs `certificatePfxBase64` and `certificatePassword`, which together are the complete mTLS credential that opens the gate; and `trigger_gsm_activation` base64s the GWeb session token into the query string at lines 530-539, so it lands in the logged URL on every GSM gate open. Lines 253-256 repeat the body at WARNING, so a failure leaks it without debug even being enabled. `README.md` lines 193-208 then tells users to turn debug logging on and attach the output to a bug report.

   Please log a length or a marker for those bodies rather than the content, and drop the token out of the `trigger_gsm_activation` URL. Lines 693 and 706 also embed the raw `get_certificate` body into a raised `CentsysApiError`, which surfaces to the user, so those want the same treatment.

Two non-blocking notes. `manifest.json` declares `paho-mqtt>=2.1` and `cryptography>=42` in `requirements`, and both ship with Home Assistant core, so declaring them risks pip resolving something that fights core's own pin. And `hacs.json` says `2024.1.0`, but `config_flow.py` line 8 imports `ConfigFlowResult`, which is 2024.4, so on 2024.1 through 2024.3 the integration fails to load at all; that floor wants to be `2024.4.0`.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

Lex-campbell:
Thanks @frenck , this was a really useful review.

So a summary of some of the changes i've made here since: 

**Logging** is fixed and shipped in 0.4.10. Rather than redact field by field, `_request` now logs only the operation, method, path and a byte count, so nothing sensitive is handed to the logger at all. That covers the token in the `trigger_gsm_activation` URL and the two `get_certificate` raises, and it also caught `CentsysApiError.__str__`, which you didn't name but which was quietly reprinting response bodies through every coordinator debug line.

**On the bearer token** - Unfortunately there's no sanctioned way to get one: no developer programme, no client registration, nothing to apply for. Its lifted from the app and its generic, not specific to my user or anything like that. Its reach is also narrower than it first looks. It's only accepted on `SendOtp` and `ValidateOtp`, so all it can do is make Centurion send a one-time code. `ValidateOtp` still needs the code that arrives on the physical handset, so it isn't an authentication bypass.

Even past that, you're not through the gate yet... Access is granted per operator by an admin/installer in the Centurion Pro app (There are two apps, Centsys Remote (the one this integration is based on) and Centsys Pro, where user management and gate parameters are set.), who has to add your number as a remote user and assign privileges (full gate opening, Pedestrian opening, time-barring etc). `GetDevicesByRemoteUserNumber` only returns operators where that's been done, so someone who completes the OTP flow on their own number just gets an empty list and no entities. The integration has no way to grant itself access, that's entirely done on the Centsys Pro app side over BLE. 

So I am fairly confident the realistic worst case is unwanted OTP messages at Centurion's expense. On that note:

**I've written to Centurion and haven't heard back yet.** I asked three things: whether they're happy with this usage, whether they'd rather issue a dedicated credential for the integration, or whether they'd consider dropping/ changing up the auth requirement on those two endpoints altogether. I'll update here as soon as they reply, happy to leave this open in the meantime as you offered.

**On the non-blocking notes:** `hacs.json` is bumped to 2024.4.0, and I checked the rest of the integration for anything needing a higher floor and found nothing. `cryptography` is dropped. I've kept `paho-mqtt` for now as it's in `package_constraints.txt` but not core `requirements.txt`, so on a Core venv without the MQTT integration it isn't installed, and `>=2.1` resolves to the pinned 2.1.0 rather than fighting it. Happy to drop it if you suggest otherwise.

Your review also prompted a wider pass of my own, which turned up a missing reauth flow (You'd have to reinstall the integration to re-auth which is ugly in 2026), background tasks surviving unload, and a telemetry loop waking unreachable gates every 60 seconds. All fixed in the same release.

PR #8968, opened 2026-07-05
Adds new integration [homeassistant-ai/ha-mcp-integration]

kingpanther13:
## Checklist

## Links

Link to current release: <https://github.com/homeassistant-ai/ha-mcp-integration/releases/tag/v0.10.0>
Link to successful HACS action (without the `ignore` key): <https://github.com/homeassistant-ai/ha-mcp-integration/actions/runs/28726303783>
Link to successful hassfest action (if integration): <https://github.com/homeassistant-ai/ha-mcp-integration/actions/runs/28726303783>

### DIFF
--- a/integration
+++ b/integration
@@ -1065,6 +1065,7 @@
   "hollowpnt92/lubelogger-ha",
   "home-assistant-HomeWhiz/home-assistant-HomeWhiz",
   "Home-Is-Where-You-Hang-Your-Hack/sensor.goveetemp_bt_hci",
+  "homeassistant-ai/ha-mcp-integration",
   "HomeAssistant-Mods/home-assistant-miele",
   "hopkins-tk/home-assistant-aseko-local",
   "hostcc/hass-gs-alarm",


frenck (commented):
Hi @kingpanther13, thanks for the submission! 👋

I am closing this, and I want to be straight with you about why, because the engineering here is not the problem. The SSRF hardening on the CIMD fetch is careful work, the websocket surface is uniformly admin-gated, and the code carries its own review annotations, which is more than most submissions manage.

What I cannot do is put this combination of defaults into the default catalog, where anyone can download it in two clicks and where the store listing is an implicit endorsement.

1. The webhook is enabled by default with auth mode `none`, so an unauthenticated `/api/webhook/<id>` proxies straight into the full MCP toolset with the random id as the only credential, and `DEFAULT_BIND_HOST = "0.0.0.0"` puts the direct port on every interface. A secret URL is a reasonable trust model for someone who chose it; it is not a reasonable default for someone who clicked Download.

2. `oauth_autoapprove.py:212` serves `/authorize` unauthenticated and, in that same default mode, 302-redirects to any spec-valid `redirect_uri` with no login and no consent. Your own docstring names the result: a crafted-link redirector on the Home Assistant origin, and it records that the exact-match allowlist shipped in July was retired on 2026-08-14. Turning a mitigation off is the direction I would need to see reversed, not documented.

3. The integration does not ship the server. On entry setup it shells out to `uv pip install` to pull `ha-mcp` from PyPI into Home Assistant's environment, with an admin-settable `pip_spec` and a six-hourly PyPI poll that can reload the entry into a new build. HACS installs what is in the repository at the release tag; an integration that fetches and swaps out its own executable payload afterwards is outside what a reviewer or a user can inspect.

4. `embedded_server.py:1158-1188` creates a Home Assistant admin user and mints a refresh token with a ten-year lifetime.

5. `embedded_setup.py:443-448` writes the OAuth client secret into the log in cleartext.

6. `install_source_check.py` and `hacs_nudge.py` read `hass.data["hacs"]` and call `update_repository(...)` and `async_update_listeners()`. Your own docstrings say there is no supported API for this, which is exactly the reason not to depend on it.

The path forward, if you want to bring it back: ship the MCP server inside the repository or as a normal `requirements` entry so what HACS installs is what runs, default the webhook to off or to `ha_auth`, bind to loopback unless the user opts out, restore the redirect allowlist, drop the secret from the log, and let Home Assistant own its own admin users. Then open a fresh PR through the publishing wizard at https://hacs.xyz/docs/publish/start. The repository works as a HACS custom repository today, which is the right home for something that asks this much trust up front.

If you would like to argue any of this with the wider HACS team rather than with me, an issue on hacs/integration is the place, and I am happy to be overruled there.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

kingpanther13:
@frenck Thank you for the review. not looking to argue, but just wanting some clarification so I know what direction to go with this so I can meet your guidelines if I choose to submit again in the future. For 1, would you consider still leaving it in "none" mode by default, but leaving WAN off by default? The webhook itself shouldn't default off because users can connect locally via the webhook as well. I feel that that would be a good compromise; ha_auth mode isn't enabled by default because of a plethora of reasons, but mainly I want users to connect without using auth initially so they can prove that their reverse proxy works at all, and also for users that are connecting only locally I feel that the extra auth isn't really necessary. For 2. This is intentional because this mode is meant to be unauthenticated other than the secret key, as above, we didn't want to overkill auth for people connecting locally or for people that chose to use just the secret key, but I understand the concern about it being the default mode. As for the allowlist, I got rid of it because I felt it would be too large of a maintenance burden and could block out legitimate servers.

For 3, it took me a minute to wrap my brain around this one but I see where your concern is. We consider the custom component and the server separate products. We often will update the server without needing to touch the custom component, especially when pushing dev versions. So just to clarify, are you suggesting that any time the server updates we should be updating the custom component in HACS as well so that they're always linked together, and remove the ability for them to update asynchronously? Is it OK to leave the ability to override the server code, just remove the automatic nature of it? Additionally just to clarify from my side, the only reason we have two repos and why everything is set up in this manner is because the initial PR I created #8966 didn't work with our main repository. I opened #8967 to try and address that.

4. I think that this is necessary for it to function; however I agree that it shouldn't be automated without user consent, would it be acceptable to have this remain but behind a user controlled flow? 
5. yeah, we can address that for sure
6.  I believe this was because HACS was slow to advertise custom component updates, and some critical updates would get out of sync; is this a bad thing for us to have? I can certainly remove it, but if we sync the server with the custom component it will be a moot point anyways I think.

PR #8996, opened 2026-07-05
Adds new integration [dsackr/ha-digital-frames]

dsackr:
## Checklist

## Links

Link to current release: https://github.com/dsackr/ha-digital-frames/releases/tag/v0.12.123
Link to successful HACS action (without the `ignore` key): https://github.com/dsackr/ha-digital-frames/actions/runs/29707719502
Link to successful hassfest action (if integration): https://github.com/dsackr/ha-digital-frames/actions/runs/29707719526

Repository was renamed from `dsackr/fraimic-homeassistant` → `dsackr/ha-digital-frames` (product: Digital Frames, domain: `digital_frames`).

### DIFF
--- a/integration
+++ b/integration
@@ -722,7 +722,7 @@
   "droans/mass_queue",
   "droidgren/smhi_season",
   "droso-hass/idfm",
-  "dsackr/fraimic-homeassistant",
+  "dsackr/ha-digital-frames",
   "dscao/ikuai",
   "DSorlov/http_agent",
   "DSorlov/snmp_printer",
--- a/integration
+++ b/integration
@@ -722,6 +722,7 @@
   "droans/mass_queue",
   "droidgren/smhi_season",
   "droso-hass/idfm",
+  "dsackr/fraimic-homeassistant",
   "dscao/ikuai",
   "DSorlov/http_agent",
   "DSorlov/snmp_printer",


frenck (changes requested):
Hi @dsackr, thanks for the submission! 👋

1. Every Samsung MDC send fails. `samsung.py` lines 95-102 declare `send_mdc_content_download(host, url, *, pin, port=..., timeout=...)`, so `pin` is keyword-only, but `samsung_coordinator.py` lines 190-195 pass `self.mdc_pin` as a third positional argument through `async_add_executor_job`, which forwards its arguments positionally. That raises `TypeError` on every call, gets caught at line 196, and surfaces to the user as "Samsung MDC send failed". Please pass it as a keyword, for example with `functools.partial(send_mdc_content_download, self.host, url, pin=self.mdc_pin)`.

2. `walls_http.py` line 244 calls `encode_for_panel_with_preview(...)` directly inside `async def post`. Every other call site in the integration wraps that function in an executor, and `panel_codec.py` documents it as an executor target. Sending a wallpaper to a wall runs one full Pillow encode per frame on the event loop, so a multi-frame wall stalls all of Home Assistant for the duration. Please wrap it in `await hass.async_add_executor_job(...)` like the other call sites.

3. Several filesystem calls run on the event loop from `async def` context. The one that matters most is `coordinator.py` lines 384-399: `get_pull_bin` does `self._pull_bin_path.read_bytes()` synchronously, and it is called bare from `async def get` at `http_api.py` lines 94-104, so every battery-frame wake pull reads the whole payload on the loop. The same pattern appears at `__init__.py` line 766 (`os.fdopen(...).write()` in `_download_media_to_temp`), `__init__.py` lines 1042 and 1072, `library.py` lines 1355 and 1363 and again at 1753 and 1761, `http_api.py` line 454, `scene_packs.py` line 571, and `update.py` line 711. Please move these through `hass.async_add_executor_job`, as the neighbouring `unlink` and `read_bytes` calls in `coordinator.py` already do.

4. `renderers/newspaper/newspaper_renderer.py` lines 446-462 catch any SSL error and retry the request with `ssl._create_unverified_context()`. The targets are fixed internet endpoints, not local devices: the RSS sources in `SOURCE_FEEDS` and, at lines 482-488, a font downloaded from `cdn.jsdelivr.net` whose bytes are then written to disk and loaded as a font file. That turns a certificate failure into a silent downgrade on a path that persists what it fetches. Please drop the unverified retry and let the fetch fail.

Not blocking, just noting for a future release: 46 of your HTTP endpoints mutate state and 6 check `is_admin`, behind a panel registered `require_admin=False` at `__init__.py` lines 422 and 441. The destructive ones (update install and restart) are gated, so this is a design call rather than a defect, but it is worth a deliberate decision. Separately, `library.py` lines 1362 and 1760 interpolate a caller-supplied `image_id` into a cache path without `_safe_image_id`, which that helper's own docstring at lines 290-296 says is required. It is not reachable today because the only caller resolves against the manifest first, but it is one new call site away from being so.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

PR #9025, opened 2026-07-06
Adds new plugin [mycrouch/ecovacs-vacuum-card]

mycrouch:
## Checklist

- [ ] (For integrations only) I've added the [hassfest action](https://developers.home-assistant.io/blog/2020/04/16/hassfest/) to my repository.

## Links

Link to current release: <https://github.com/mycrouch/ecovacs-vacuum-card/releases/tag/v1.3.1>
Link to successful HACS action (without the `ignore` key): <https://github.com/mycrouch/ecovacs-vacuum-card/actions/runs/28795998522>

### DIFF
--- a/plugin
+++ b/plugin
@@ -455,6 +455,7 @@
   "mtheli/toothbrush-card",
   "MUbeira0/lovelace-vigobus-card",
   "mutilator/homeassistant-solar-panel-preview",
+  "mycrouch/ecovacs-vacuum-card",
   "naked-head/puffer-card",
   "nathan-gs/ha-map-card",
   "nathanmarlor/foxess_modbus_charge_period_card",
--- a/plugin
+++ b/plugin
@@ -478,6 +478,7 @@
   "mtheli/toothbrush-card",
   "MUbeira0/lovelace-vigobus-card",
   "mutilator/homeassistant-solar-panel-preview",
+  "mycrouch/ecovacs-vacuum-card",
   "naked-head/lektrico-charger-card",
   "naked-head/puffer-card",
   "nathan-gs/ha-map-card",


mycrouch:
Withdrawing this submission for now — apologies for the noise.

frenck (changes requested):
Hi @mycrouch, thanks for the submission! 👋

Same escaping point as on the AirTouch card, and the surface here is a bit wider. In `ecovacs-vacuum-card.js`, freeform values reach `innerHTML` unescaped at:

- `:328`, where the gradient string from your card config is interpolated directly into a quoted `style=` attribute.
- `:632` (`fan_speed_list` values), `:669` and `:892` (keys of the `rooms` state attribute).
- `:872` and `:873`, the user-typed schedule name and the room-derived summary.
- `:919`, which renders `_schedStatus`, and that can carry `e.message` text from a failed Home Assistant call (set at `:1246`).
- `:1408`, `:1434`, `:1442`, the theme gradient value and installed theme names.

`:899` shows you already had the problem in mind: it escapes `"` in the schedule-name input, but only `"`. An `esc()` helper applied consistently, or `textContent` for the text-position cases, is the clean fix. Entity IDs need no escaping.

Separately, `:1619-1621` call `customElements.define` and `:1623` pushes to `window.customCards` without the `customElements.get` guard you used on the AirTouch card, so a double-load of the resource throws and duplicates the theme-picker entry. Worth matching the two.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

mycrouch:
Thanks, that's all done in v1.5.2.

Everything on your list now goes through an `evcEsc()` helper, including the schedule name input where the quote-only escape was (that's gone). I also caught the entity id in the not-found message, and the room ids and time value on the schedule tiles while I was in there. The `customElements.define` calls and the `customCards` push are guarded now the same way as the AirTouch card, and I've tested a double load doesn't throw or double up the entry.

Can't re-request review from my end, so flagging it here. @frenck

mycrouch:
Correction on the double-load claim above: the guards work, but they weren't the whole story. Loaded a second time as a classic script the file throws on the top-level `const` declarations before it ever reaches the guards, so my test was only proving half of it. v1.5.3 wraps the file in an IIFE, which fixes that and keeps the module-scope names out of the global scope. Please review v1.5.3 rather than v1.5.2.

PR #9052, opened 2026-07-08
Adds new integration [NiklasM-foss/ha-huellhorst-abfuhr]

NiklasM-foss:
## Checklist

## Links

Link to current release: https://github.com/NiklasM-foss/ha-huellhorst-abfuhr/releases/tag/v1.0.0
Link to successful HACS action (without the `ignore` key): https://github.com/NiklasM-foss/ha-huellhorst-abfuhr/actions/runs/28924659821/job/85809393558
Link to successful hassfest action (if integration): https://github.com/NiklasM-foss/ha-huellhorst-abfuhr/actions/runs/28924659821/job/85809393550


frenck (changes requested):
Hi @NiklasM-foss, thanks for the submission! 👋

1. `custom_components/huellhorst_abfuhr/api.py` lines 159-162 do a function-local `from zoneinfo import ZoneInfo` and then construct `ZoneInfo("Europe/Berlin")`, and both of those touch the disk: the import machinery on first call, and the tz database read for an uncached zone. That code sits in `_parse_ics_datetime`, a plain `def`, reached from `parse_ics` at `coordinator.py` line 66, which runs inside `async def _async_update_data` with no executor. It happens once per VEVENT, so a refresh with a full year of collection dates does that work dozens of times on the event loop. Home Assistant provides `homeassistant.util.dt.async_get_time_zone` for exactly this. While you are there, the zone is hardcoded rather than read from `hass.config.time_zone`, so a user running Home Assistant in another zone gets times shifted against the rest of their system.

2. `hacs.json` line 3 declares `"homeassistant": "2024.4.0"`, but the integration ships brand assets in-tree at `custom_components/huellhorst_abfuhr/brand/`, and Home Assistant only reads those from 2026.3 onwards. On an older core the integration shows no icon and there is nothing to tell the user why. Please raise the floor to `"2026.3.0"`.

Not blocking: `sensor.py` line 68 and line 97 and `calendar.py` line 41 set an explicit `_attr_name` alongside `_attr_has_entity_name = True`, and the explicit name wins, so the `_attr_translation_key` values never take effect. Neither `strings.json` nor either translations file carries an `entity` block, so `translations/en.json` currently cannot change an entity name. Worth wiring up if you want the English names to work. Also, `sensor.py` line 98 builds a translation key from whatever slug `_slug_for_summary` invents for an unrecognised ICS summary, and those have no matching entry in `icons.json` or the translations.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

NiklasM-foss:
Thanks for the review, @frenck. Both points are addressed in [v1.1.0](https://github.com/NiklasM-foss/ha-huellhorst-abfuhr/releases/tag/v1.1.0).

- The time zone is no longer resolved inside `_parse_ics_datetime`. The coordinator now fetches it once per refresh with `dt_util.async_get_time_zone(self.hass.config.time_zone)` and passes it down into `parse_ics()`, so no import machinery or tz-database read happens on the event loop, and the zone follows the user's Home Assistant configuration instead of a hardcoded `Europe/Berlin`. Date-only values are anchored to local midnight as well now.
- `hacs.json` requires `2026.3.0`.

I also took the non-blocking note: the explicit `_attr_name` values are gone from both sensor classes and the calendar, and `strings.json` plus both translation files now carry an `entity` block with the names. The calendar had no `translation_key` at all, it has one now. A slug coming from an unrecognised ICS summary no longer gets an invented `translation_key`; it keeps an explicit name and icon, since there is no translation or `icons.json` entry for it.

Validation is green. Could you re-request review when you get a chance?

PR #9086, opened 2026-07-09
Adds new integration [DeKaN/ha-syncleo]

DeKaN:
## Checklist

## Links

Link to current release: https://github.com/DeKaN/ha-syncleo/releases/tag/0.1.0
Link to successful HACS action (without the `ignore` key): https://github.com/DeKaN/ha-syncleo/actions/runs/29047476833/job/86219608347
Link to successful hassfest action (if integration): https://github.com/DeKaN/ha-syncleo/actions/runs/29047476833/job/86219608333


frenck (changes requested):
Hi @DeKaN, thanks for the submission! 👋

The device authorization token ends up in the Home Assistant log at INFO level, so it lands in `home-assistant.log` with default logging settings, not just when someone turns on debug.

`custom_components/syncleo/config_flow.py` lines 205-209 log the whole entry dict:

```
_LOGGER.info(
    "Creating config entry %s with data %s",
    entry_title,
    self._entry_data,
)
```

`self._entry_data` carries `CONF_TOKEN` and `CONF_PUBLIC_KEY` (set at line 165 from `utils.parse_share_url`), so the Share URL token is written out verbatim. The same value also appears at DEBUG on lines 127 and 171-175, which is less exposed but the same secret. Logs get pasted into issue reports, so please log the host and title only, or redact the token before logging.

Two smaller things while you are in there. `entity.py` lines 169, 180 and 242 log every inbound and outbound UDP command at INFO, which is noisy at the default level for a device that pushes continuously; DEBUG would suit those better. And `devices/cookers.py` ships as a zero-byte file inside `syncleo.zip`.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

DeKaN:
Hi @frenck, thanks for review.

Device public key is exposed via mdns by device (and it regenerates after every power loss), so I think we can log it without masking. I'll add masking for device token.

The integration now is in active development, so logging commands at level info helps at this moment, because debug provides much more logs with raw data. I'm planning to move this logging to debug after adding support for some "hard" Polaris devices (coffee machines, aerogrils and vacuums).

DeKaN:
I've added masking for token and removed empty `devices/cookers.py` in release 0.6.6.

frenck (changes requested):
Hi @DeKaN, thanks for masking the token and dropping the empty cookers module! 👋

1. The command logging in `custom_components/syncleo/entity.py` is still at INFO: lines 172 (`Sending command %s`), 183 (`Also sending command %s`) and 245-249 (`Start handling update for device ...`). For a device that pushes continuously this fills the default log. Please move those three to DEBUG.

2. `custom_components/syncleo/select.py` line 127 raises `UnboundLocalError` on every unchanged data-source report. In `_handle_device_update`, the data-source branch at lines 93-103 only returns when the reported MAC differs from `self._current_option`. In the steady state the heater re-reports the same data source on each push, so control falls through to `text_value = self._rev_options_map.get(value)` with `value` never assigned. This reaches `HeaterProfile` through `FEATURE_DATA_SOURCE_CURRENT_TEMPERATURE` (`devices/heaters.py` line 348). That branch needs an unconditional return.

3. `_parse_temp` in `custom_components/syncleo/broadcaster.py` (lines 107-117) returns `0` when the source sensor is unavailable, unknown or non-numeric, and that value goes straight into the mDNS TXT record at lines 62 and 91. At startup before the sensor has a state, the integration advertises a fabricated 0 degrees to devices that use it as their temperature data source, so a heater in anti-frost or eco mode drives to full output. Please keep the last known value or omit the property rather than substituting zero.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

DeKaN:
Hi @frenck, thanks for good review.

I've fixed found issues and rewrote code in broadcaster (not only for zero temperature)

frenck (changes requested):
Hi @DeKaN, thanks for all three, and the broadcaster fix is exactly right: unregistering the service rather than advertising a fabricated zero means a heater falls back to its own sensor instead of driving to full output, and a genuine 0.0 reading still publishes because the guard tests `is not None` rather than truthiness. 👋

Two things, both the same shapes one file over from where they were fixed.

1. The per-push INFO logging is still there, in the platform subclasses rather than the base class. `entity.py` lines 172, 183 and 245 are DEBUG now, but the subclasses that run on the same callback are not: `binary_sensor.py:54`, `climate.py:126` and `:176`, `fan.py:99`, `number.py:83` and `:92`, `select.py:107` and `:115`, `sensor.py:69` and `:77`, `switch.py:58` and `:66`, and `water_heater.py:90` are all INFO on the device-push path. Four of them are worse than per-match: `number.py:69`, `switch.py:56`, `select.py:105` and `sensor.py:67` test `self._is_program_data`, a flag set in `__init__`, rather than testing the incoming command, so they log on every inbound command of any type. On a rusclimate type 46 heater with two program-data numbers, that is several INFO lines per push at the device's native rate. `climate.py:126` also logs on every MODE push whether or not the mode changed. The default log fills as before; the three lines that moved were the minority of the volume.

2. `number.py` raises `UnboundLocalError` in `__init__` for the rusclimate type 14 heater, and it takes the whole number platform down with it. Lines 40 to 45 bind `config` in the `if self._is_program_data` branch or the `elif ... in profile.number_configs` branch, then read it at line 45; neither binds when a key listed in `numbers` is in neither mapping. `devices/heaters.py:168` declares `numbers=[FEATURE_POWER_LEVEL, PD_ECO_DELTA, PD_ANTI_FROST_TEMP]` while `program_data_fields` holds only `PD_POWER` and `number_configs` only `FEATURE_POWER_LEVEL`, so `PD_ECO_DELTA` matches neither. Since `async_setup_entry` builds the entities in a list comprehension, the exception escapes and the platform never sets up, so that device loses its power-level number too. The `_LOGGER.warning("No number_config found for %s, using defaults")` at line 50 is unreachable, because the failure it exists to handle raises one line earlier. A `config = None` initialiser before line 40 makes that warning live; adding the two missing entries to the profile fixes it properly. I checked every other profile's `numbers` list against its own mappings and this is the only one with the mismatch.

One non-blocking note: the README does not mention the virtual temperature broadcaster at all, which is the feature this release exists to fix and the one with physical consequences, since it feeds a heater's temperature data source. The add-integration flow that picks the source sensor is worth documenting.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

DeKaN:
Hi @frenck, thanks for finding copy-paste bug with device type 14. I've fixed it in both places: in `number.py` and in `heaters.py`.
Also, I decreased log level in mentioned places and added some info about virtual temperature sensor in README

PR #9089, opened 2026-07-10
Adds new integration [whoamib2/homeseer-home-assistant-bridge]

whoamib2:
## Checklist

- [x ] I've read the [publishing documentation](https://hacs.xyz/docs/publish/start).
- [x ] I've added the [HACS action](https://hacs.xyz/docs/publish/action) to my repository.
- [x ] (For integrations only) I've added the [hassfest action](https://developers.home-assistant.io/blog/2020/04/16/hassfest/) to my repository.
- [ x] The actions are passing without any disabled checks in my repository.
- [ x] I've added a link to the action run on my repository below in the links section.
- [x ] I've created a new release of the repository after the validation actions were run successfully.

## Links

Link to current release: <https://github.com/whoamib2/homeseer-home-assistant-bridge/releases/tag/v3.8.1>
Link to successful HACS action (without the `ignore` key): <https://github.com/whoamib2/homeseer-home-assistant-bridge/actions/runs/29063448397>
Link to successful hassfest action (if integration): <https://github.com/whoamib2/homeseer-home-assistant-bridge/actions/runs/29062815142>


frenck (changes requested):
Hi @whoamib2, thanks for the submission! 👋

There is a lot of real work here: seven platforms, mcsMQTT push, diagnostics with redaction, area suggestions with a dry run by default. The blocker is one specific thing.

1. The integration writes Home Assistant's own storage. `custom_components/homeseer_bridge/dashboard.py` line 80 opens `Store(hass, 1, "lovelace_dashboards")`, which is the `lovelace` component's registry file, loads it, appends its own entry at lines 93-103, and saves it back at line 110; line 105 also claims the `lovelace.homeseer-bridge` key in core's dashboard-config namespace. `__init__.py` line 338 fires this unconditionally on every setup, with no way for a user to decline. The existence guard at lines 84-91 means you are not clobbering an existing dashboard, which is good, but writing that file directly still bypasses the `lovelace` component entirely: the running frontend's collection never learns about the new dashboard, so the sidebar entry only appears after a restart, and nothing coordinates the write with core or with a user who is editing dashboards at that moment. An integration should not edit another component's storage.

You already have the right version of this feature: the `create_dashboard` service in `services.yaml`. Please make that the only path, so the user asks for the dashboard rather than getting it silently, and build it through the `lovelace` component's collection API rather than a raw `Store`.

2. `const.py` line 25 sets `DEFAULT_VIRTUAL_POLL_INTERVAL_SECONDS = 5`, wired up at `__init__.py` lines 429-432. Each tick calls `api.async_get_status()`, which is the full-installation `/JSON?request=getstatus` fetch, and then re-normalises every ref. That is a whole-house HTTP round trip and reparse every five seconds, on an integration whose README specifically targets large HomeSeer installations. Given mcsMQTT already provides the live updates, please raise that default substantially; 30 seconds or more would still be responsive for the virtual-device fallback.

3. Your own installation's values ship as defaults for everyone. `const.py` line 7 has `DEFAULT_HS_URL = "http://192.168.0.193"`, line 8 `DEFAULT_MQTT_PREFIX = "Homeseer/Chip23/mcsMQTT"`, line 30 `DEFAULT_ACTIVITY_EXCLUDED_TERMS = "utility,jon00"`, and line 9 an excluded-terms list naming specific vendors. These are prefilled into the config flow at `config_flow.py` lines 42-45, so every new user starts with your controller name and your filters. Please blank them or use neutral placeholders.

One more for a future release, not blocking: the README opens with the changelog and buries the description about 4000 bytes down, so someone landing on the repo has to scroll a long way to find out what it does.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

frenck (changes requested):
Hi @whoamib2, all four are done, and the dashboard rework is the right shape: no `Store` anywhere in the integration now, creation goes through `DashboardsCollection` and `LovelaceStorage` with `frontend.async_register_built_in_panel`, so the sidebar entry appears without a restart, and it is reachable only from the `create_dashboard` service instead of firing on every setup. The poll default is 30 seconds, the four personal defaults are blanked with `vol.Length(min=1)` behind them, and the README leads with the description. 👋

Two things.

1. `topic_map.py` ships your own installation as a hardcoded table, and it takes priority over the user's own topics. `REF_TO_TOPICS` is a 9.4 KB literal of roughly 130 HomeSeer refs mapped to `Homeseer/Chip23/mcsMQTT/...` topics, and `helpers.py` lines 74 to 82 consume it first, before the generated candidates: line 80 assigns the full topic key outright, and line 82 `setdefault`s a name-only key `__name__/<last segment>` that is claimed before the user's own devices reach line 90. Several of those last segments are generic: `wled/garage` gives `__name__/garage`, `.../1659_4-Plex/Switch` gives `__name__/switch`, and there are `__name__/window` and `__name__/api` entries too. Ref numbers are small integers, so on a large HomeSeer install a collision is likely, and `topic_to_ref` falls back to exactly that key. An incoming MQTT message on any topic ending in `/Switch` or `/Window` that does not match a full topic then routes to the wrong ref, and `apply_mqtt_state` writes a foreign device's value into that entity. For a table that includes door locks and water valves, mis-routed state is a correctness problem worth closing rather than a cosmetic one. Emptying `REF_TO_TOPICS` and relying on `topic_candidates`, or making the exact map a user-supplied option, both work.

   The same file is also the item-3 problem at a much larger scale: it discloses your floor plan, room names, children's first names, tenant unit names and a third-party account identifier to everyone who installs the integration. That is your call to make about your own repository, and I am not asking you to change anything for that reason alone, but it is worth knowing that it ships to every user.

2. The declared floor is well below what the code needs, and it fails in two different ways below it. `hacs.json` says `2025.1.0`. `__init__.py` lines 126 to 129 call `DeviceRegistry.async_get_device_by_identifier(identifier, config_entry_id)`, which arrived in 2026.8.0 (2026.7.0's `device_registry.py` has only `async_get_device`), so on anything older the first exclusion-bearing setup raises `AttributeError`. `dashboard.py` line 118 calls `frontend.async_panel_exists`, added in 2026.5.0. And `__init__.py` line 44 imports `LOVELACE_DATA` from `lovelace.const`, which does not exist before 2025.2.0, so on 2025.1.x the integration fails to load at all with an `ImportError`. Please set `"homeassistant": "2026.8.0"`. If you would rather keep the wider range, `async_get_device({(DOMAIN, str(ref))})` and a `DATA_PANELS` membership test are the long-standing equivalents, and then `2025.2.0` covers you.

Four non-blocking notes for the same release. `manifest.json` now makes `lovelace` a hard `dependency` for a dashboard created only on explicit request, where `after_dependencies` states the real relationship and `dashboard.py` lines 93 to 96 already handle its absence. All five services are registered with no admin check, and `apply_suggested_areas` mutates the area and device registries while `create_dashboard` adds a sidebar item for everyone. `diagnostics.py` line 14 does not redact `homeseer_url`, and since there is no credential field the only way to authenticate is `http://user:pass@host` in that URL, which would then land verbatim in a downloadable diagnostics file. And `light.py` lines 76 to 79 and `fan.py` lines 69 and 77 do `float(...)` and `> 0` on the result of `on_value`, which returns the raw CAPI field when it is not parseable as a float, so a non-numeric control value raises out of `async_turn_on`; `switch.py` lines 70 to 73 already guard that.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

PR #9093, opened 2026-07-10
Adds new integration [m-hume/tasmota_timers]

m-hume:
## Checklist

## Links

Link to current release: https://github.com/m-hume/tasmota_timers/releases/tag/v0.3.0
Link to successful HACS action (without the `ignore` key): https://github.com/m-hume/tasmota_timers/actions/runs/29083035610/job/86330121763
Link to successful hassfest action (if integration): https://github.com/m-hume/tasmota_timers/actions/runs/29083035610/job/86330121747


frenck (changes requested):
Hi @m-hume, thanks for the submission! 👋

1. `hacs.json` declares `"homeassistant": "2023.1.0"`, but `custom_components/tasmota_timers/config_flow.py` lines 36-38 return `TasmotaTimersOptionsFlow()` with no arguments and line 57 reads `self.config_entry.options`. The implicit `OptionsFlow.config_entry` property only exists from Home Assistant 2024.12.0. On anything between 2023.1 and 2024.11, which your floor explicitly permits, pressing Configure raises `AttributeError` and the dialog dies with an unknown error, so those users can never add the manual device topics your README tells them to use when discovery fails. Please set `"homeassistant": "2024.12.0"`.

2. `custom_components/tasmota_timers/__init__.py` lines 166-170 make three `await mqtt.async_subscribe(...)` calls and discard every returned unsubscribe callable, and `async_unload_entry` at lines 292-306 never unsubscribes. Reloading the integration, or removing it entirely, leaves those wildcard subscriptions live for the rest of the Home Assistant process. Each reload adds another set, so every discovery message gets handled repeatedly, and the stale closures call `async_update_entry` on a torn-down entry and push entities through the previous platform's `async_add_entities`, which raises. Please wrap each as `entry.async_on_unload(await mqtt.async_subscribe(...))`.

3. `__init__.py` lines 174-178 read `entry.options.get(CONF_DEVICE_TOPICS, "")` once during setup, and nothing registers an update listener. A user follows the Manual Device Topics steps in your README, enters their topics, saves, and sees no new entities, with no error and nothing in the log, until they reload the entry or restart. Please register `entry.async_on_unload(entry.add_update_listener(...))` with a listener that calls `hass.config_entries.async_reload(entry.entry_id)`.

Four smaller things worth folding in: add `"integration_type": "service"` to `manifest.json`; either set `"single_config_entry": true` or call `async_set_unique_id` and `_abort_if_unique_id_configured`, which would also make the `already_configured` string in `strings.json` reachable, and move the `hass.services.async_remove` block inside the `if unload_ok` branch so unloading one entry does not strip the services from another; complete `brand/` with `icon@2x.png` and the two logo files; and consider checking for a Tasmota-shaped payload before creating entities from `tele/+/LWT` and `tele/+/STATE`, since any non-Tasmota MQTT device publishing under `tele/` currently gets a switch entity persisted into the config entry with no way to remove it.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

m-hume:
Hi @frenck, all feedback has been addressed in release v0.3.0 (https://github.com/m-hume/tasmota_timers/releases/tag/v0.3.0). Could you please re-review? Thank you!

frenck (changes requested):
Hi @m-hume, thanks for all three, and for the extras: the floor is `2024.12.0`, every `mqtt.async_subscribe` is now wrapped in `entry.async_on_unload`, the update listener is registered and released properly, `integration_type` is set, the unique-id abort makes `already_configured` reachable, the service removal moved inside `if unload_ok`, and `brand/` is complete. 👋

Three things, and the first is a side effect of the item-3 fix rather than something you had before.

1. The update listener and `async_update_entry` form a reload loop during discovery. `_save_devices` at `__init__.py` lines 96 to 107 persists newly discovered devices with `hass.config_entries.async_update_entry(entry, data=new_data)`, and Home Assistant fires entry update listeners on any change made through that call, not only options changes. So on a fresh install with retained `tasmota/discovery/+/config` messages: device 1 is persisted, `entry.data` changes, `_update_listener` fires, the whole entry reloads (entities removed, all six services deregistered, subscriptions dropped), the retained discovery replays, device 2 is persisted, and so on. It does converge, because once a device is in `entry.data` the early return at line 113 stops the write, so this is not an infinite loop. But it is one full integration reload per newly discovered device, forever, including each time a new Tasmota device is flashed onto the network, with switches flapping to unavailable and `ServiceNotFound` for any `tasmota_timers.*` call that lands in the unload window. Comparing the options before reloading, or moving device persistence out of `entry.data` into a `helpers.storage.Store`, both fix it.

2. Persisted devices can never be removed. Devices are only ever added: lines 116 to 120 write into `entry_data["devices"]` and lines 98 to 107 serialise the whole dict back, and there is no deletion path anywhere in the package. Someone who adds a topic under Manual Device Topics and then deletes it from the options flow keeps the switch, recreated on every restart, and the only way out is deleting and re-adding the config entry. This is the half of the last round's final item that is still open.

3. The LWT discovery gate is not a Tasmota-shape check. The `/STATE` branch at lines 170 to 181 is a real one, testing for `Uptime`, `Heap`, `Power` or `Wifi` in the payload. The `/LWT` branch at lines 167 to 169 only tests `msg.payload != "Online"`, and `Online` is a generic MQTT last-will convention, so any non-Tasmota client publishing `tele/<name>/LWT` with `Online` still gets a switch created and persisted, which combined with item 2 cannot then be removed. Requiring a subsequent Tasmota-shaped `STATE` or a discovery message before persisting would close it.

Two non-blocking notes for the same release. The brand icons are one step undersized against the spec: `icon.png` is 128 by 128 where 256 is expected and `icon@2x.png` is 256 where 512 is; the two logos are correct. And `LICENSE` reads "Copyright (c) 2026" with no holder name filled in.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

PR #9110, opened 2026-07-11
Adds new integration [snowmonkeylab/ha-vinyl-collection]

snowmonkeylab:
## Checklist

## Links

Link to current release: https://github.com/snowmonkeylab/ha-vinyl-collection/releases/tag/v0.3.7
Link to successful HACS action (without the `ignore` key): https://github.com/snowmonkeylab/ha-vinyl-collection/actions/runs/29135468007
Link to successful hassfest action (if integration): https://github.com/snowmonkeylab/ha-vinyl-collection/actions/runs/29136636160


frenck (changes requested):
Hi @snowmonkeylab, thanks for the submission! 👋

1. `custom_components/vinyl_collection/icon.png` sits in the component root rather than in `custom_components/vinyl_collection/brand/`, which since Home Assistant 2026.3 is the only in-tree path that gets read. It is also a stale copy: its bytes differ from the current `brand/icon.png`, left over from the commit that replaced `hacs.png` with the new record icon. Please delete it. The unreferenced `hacs.png` at the repository root is byte-identical to `brand/icon.png` and can go at the same time.

2. `hacs.json` line 5 declares `"homeassistant": "2024.1.0"`, but `config_flow.py` lines 45-50 return `VinylCollectionOptionsFlow()` with no arguments and the flow then reads the implicit `self.config_entry` at lines 63, 64, 66-68 and 70-72. That property only exists from Home Assistant 2024.12.0. On 2024.1 through 2024.11, which your floor explicitly invites, pressing Configure raises `AttributeError` at line 63 and the dialog dies with an unknown error, so the Discogs token can never be set and search and cover art can never be enabled. Please set `"homeassistant": "2024.12.0"`.

3. Services outlive a reload and keep writing to an orphaned store. `__init__.py` line 108 does `hass.data.setdefault(DOMAIN, {})` and line 119 puts the non-entry key `_frontend_registered` into that same dict. On unload, line 174 pops the entry id but the `if not hass.data[DOMAIN]:` test at line 175 is never true, because that flag keeps the dict truthy, so the service-removal loop at lines 176-185 never runs. On the next setup the `has_service` guard at line 164 is already true, `_register_services` at line 165 is skipped, and the still-registered handlers keep their closure over the previous `VinylCollectionStore` from line 152 while the sensors bind to the new one at `sensor.py:20`. After a routine Reload, every `add_record` and `remove_record` writes into the orphaned store while both count sensors read the new one, so they freeze at their post-reload values and never move again until Home Assistant is fully restarted, and automations keyed on those counts silently stop firing. Please keep the frontend flag out of the per-entry dict, or exclude it from the emptiness test, and look the store up per entry rather than capturing it once.

4. `__init__.py` lines 237-248: `handle_update_record` fires `EVENT_RECORD_UPDATED` and returns without calling `_refresh_sensors(hass)`, unlike `handle_add_record` at line 221 and `handle_remove_record` at line 234. Calling `vinyl_collection.update_record` with `is_wishlist: true` therefore moves the record in storage while neither counter changes, and they stay wrong until some other record happens to be added or removed. A script that bulk-flags records to the wish list leaves both counts silently incorrect.

Three smaller things: `manifest.json` line 12 says `"version": "0.3.0"` against release tag `v0.3.7`; `iot_class` is `local_polling` where nothing polls (`sensor.py:32` sets `_attr_should_poll = False`) and the Discogs lookups are cloud calls, so `local_push` fits better; and `__init__.py` lines 121-145 auto-register the card by reaching into `hass.data["lovelace"]` internals behind a bare `except Exception`, so when that private API changes the card silently stops registering with only a warning in the log.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

frenck (changes requested):
Hi @snowmonkeylab, thanks for all four: both stray icons are gone, the floor is `2024.12.0`, the frontend flag moved out of `hass.data[DOMAIN]` so the emptiness test actually fires and the services get torn down on unload, `_store()` resolves per call instead of closing over one instance, and `handle_update_record` now refreshes the sensors. I traced the reload path rather than taking the release notes for it, and it is genuinely closed. 👋

One thing, and it is in a file this release did not touch, so I missed it last round rather than you introducing it.

`vinyl-collection-card.js` lines 1352 to 1355 define `_esc` as a `textContent` round trip:

```js
_esc(str) {
  const d = document.createElement("div");
  d.textContent = String(str != null ? str : "");
  return d.innerHTML;
}
```

Text-node serialization escapes `&`, `<` and `>`, but it leaves a double quote untouched. That output then goes inside a quoted attribute at line 246:

```js
return "<img src=\"" + this._esc(url) + "\" width=\"" + size + "\" ... onerror=\"...\"/>"
```

`url` there is `r.cover_url`, taken verbatim from the Discogs search response at `__init__.py` line 343 and rendered on every dashboard load at lines 1028, 1065, 450, 364 and 336. A cover URL containing a double quote closes `src` early and the rest is parsed as further attributes, so `https://host/a.jpg" onerror="…` gives an attacker-chosen `onerror` that wins over the built-in one, executing in the dashboard viewer's session with full frontend privileges. Shadow DOM does not isolate that.

Two producers reach it, and the second is the one that decides the severity. Discogs is a vendor feed outside the dashboard author's control. But `vinyl_collection.add_record` and `update_record` also accept `cover_url` as a bare `cv.string` with no URL validation (`__init__.py` lines 73 and 93) and no admin gate, so any authenticated non-admin user or script can store the payload and have it run in an admin's browser next time they open the dashboard. The same exposure exists at line 1265 for the Spotify `thumbnail`. Adding `.replace(/"/g, "&quot;")` to `_esc` closes it; setting `img.src` as a DOM property instead of building the string, plus validating `cover_url` as an http or https URL server-side, closes it properly.

Three non-blocking notes for the same release. `iot_class` moved from `local_polling` to `local_push`, and `local_polling` was indeed wrong, but nothing pushes either: the only network I/O is on-demand calls to `api.discogs.com`, so `cloud_polling` is the honest value. The `resource_type` schema default is `"master"` while `services.yaml` documents `"release"` and the handler's `"release"` fallback is unreachable. And the README's services table omits `get_discogs_tracks`, which is registered and documented in `services.yaml`.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

frenck (approved):
Hi @snowmonkeylab, thanks for escaping the quote in _esc and validating cover_url server-side! 👋

Approving and merging now.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

PR #9123, opened 2026-07-12
Adds new integration [ummeco/praycalc-ha]

acamarata:
## Checklist

## Links

Link to current release: <https://github.com/ummeco/praycalc-ha/releases/tag/v0.7.1>
Link to successful HACS action (without the `ignore` key): <https://github.com/ummeco/praycalc-ha/actions/runs/29174915617>
Link to successful hassfest action (if integration): <https://github.com/ummeco/praycalc-ha/actions/runs/29174915617>

### DIFF
--- a/integration
+++ b/integration
@@ -2546,6 +2546,7 @@
   "ufozone/ha-zcs-mower",
   "UI-Lovelace-Minimalist/UI",
   "uiotlink/ha_uiot_home",
+  "ummeco/praycalc-ha",
   "ungiglio/ha-xev-yoyo",
   "unii-security/homeassistant-unii",
   "UnoSite/IsItPayday",


frenck (changes requested):
Hi @acamarata, thanks for the submission! 👋

1. There is no `custom_components/praycalc/translations/` directory, so every entity name and every config-flow string renders untranslated. Home Assistant loads runtime translations only from `<integration>/translations/<language>.json`; `strings.json` is a build-time source file that core reads at build time and a custom integration never loads at runtime. All nine entities set `_attr_has_entity_name = True` together with a `_attr_translation_key` (`sensor.py:135` for `next_prayer`, `:235` for the six prayer keys, `:285` for `qibla`, `:329` for `hijri_date`), so with nothing to resolve, each falls back to an empty name: the user gets nine entities that all present as the bare device name with `_2` through `_9` suffixes on the generated entity IDs. That also means the entity IDs your README documents (`sensor.praycalc_fajr`, `sensor.praycalc_qibla` and the rest) are not what Home Assistant actually creates, so the automation examples do not work as written. The config flow is affected too: the step title, the field labels and descriptions, and the `cannot_connect`, `invalid_latitude` and `invalid_longitude` messages all render as raw keys. Please copy `strings.json` to `custom_components/praycalc/translations/en.json`, keeping `strings.json` as the source. Worth knowing that hassfest does not validate custom-integration translations, which is why this passed CI.

2. `hacs.json` line 3 declares `"homeassistant": "2024.1.0"`, but `config_flow.py` line 10 imports `ConfigFlowResult`, which Home Assistant added in 2024.4.0. A user on 2024.1 through 2024.3 passes the HACS version gate, installs, and the integration fails to load with `ImportError: cannot import name 'ConfigFlowResult'`, leaving a setup-failed entry, no entities and a traceback in the log. Please raise it to at least `"2024.4.0"`. `"2026.3.0"` is worth considering instead, since your `custom_components/praycalc/brand/` assets are only served from 2026.3 and below that the integration shows no icon at all.

Some smaller things worth folding into the same release. `coordinator.py` lines 15 and 55 set a 60-second update interval against a third-party cloud endpoint, which is 1440 requests a day per install for data that changes once a day; a 15 to 30 minute interval with the countdown recomputed locally would be much kinder to `smart.praycalc.com`. `sensor.py` lines 250-260 build the `TIMESTAMP` value with `datetime.now().replace(...).astimezone()`, which uses the Home Assistant host's clock rather than the timezone of the configured coordinates, so it drifts by the offset difference for anyone configuring a location elsewhere; `homeassistant.util.dt` is the right tool. `sensor.py` lines 147-181 assign `_attr_native_value` and `_attr_extra_state_attributes` in `_update_state()` while the class also overrides both as properties, so that half of the method is dead code. `sensor.py` line 73 passes `update_before_add=True` on `CoordinatorEntity` subclasses, which adds a redundant refresh per entity on top of `async_config_entry_first_refresh()`. And there is no options or reconfigure flow, so changing the calculation method or madhab means deleting and re-adding the entry.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

acamarata:
Thanks for the review, @frenck. Every point was correct, and chasing them turned up three more bugs that were worse than what the code review could show. All of it is in [v0.8.0](https://github.com/ummeco/praycalc-ha/releases/tag/v0.8.0).

### The two blockers

**Translations.** Added `custom_components/praycalc/translations/en.json`, keeping `strings.json` as the source. You were right that this was invisible to CI, so I added a job that fails if a key in `strings.json` is missing from `translations/en.json`, and a second that fails if any `_attr_translation_key` in the code has no matching entry.

**Version floor.** Went with your stronger suggestion, `2026.3.0`, rather than the 2024.4.0 minimum. There is no installed base to strand yet, and shipping an integration whose icon silently does not render is not worth the extra version breadth.

### Entity IDs, and why the README was worse than you thought

You asked me to re-verify the documented IDs. I did that by booting Home Assistant 2026.9.0b1 in a test harness and reading the entity registry, rather than by reasoning about it. Two things were wrong:

- `sensor.praycalc_qibla` is actually `sensor.praycalc_qibla_direction`. The translated name is "Qibla direction", which is what the ID slug is built from.
- Every ID carries the configured city. The city field is required and pre-filled from `hass.config.location_name`, so a typical user gets `sensor.praycalc_detroit_fajr`, not `sensor.praycalc_fajr`.

So the automation examples were wrong for essentially everyone. The README now documents the blank-city form, states the city rule up front, and points at the entities page. Those IDs are asserted in tests so the docs cannot drift again.

### The smaller items

All five are in, and the coordinator interval was the right one to prioritise. It is 30 minutes now, so 48 requests a day per install instead of 1440. The countdown still ticks every minute, recomputed locally from cached data with no network call. `homeassistant.util.dt` throughout, dead assignments in `_update_state()` gone, `update_before_add` gone. There is now an options flow for method, madhab and API URL, and a reconfigure flow for location, so neither needs a delete and re-add.

### Three things you could not have seen from the diff

Verifying your timezone point against the live API turned up worse:

1. **The Hijri sensor was raising `AttributeError` on every state update.** The API returns `hijriDate` as a formatted string, and the code called `.get()` on it as though it were a mapping.

2. **Every prayer time was wrong by the local UTC offset.** Your read was that it drifts for remote locations. It was actually worse: the API returns UTC wall-clock strings with no date attached, and the integration treated them as local wall-clock, so the times were wrong for everyone. They are now anchored to real instants, rolled forward across UTC midnight, then re-anchored so Dhuhr falls on the requested local date. Checked against eight cities from UTC-8 to UTC+14.

3. **The backing API dropped Fajr and Sunrise everywhere east of Greenwich.** Its formatter returned `--:--` for any hour that came out negative, which is exactly where those two land for Indonesia, Malaysia, Japan, Australia and New Zealand. A separate rounding bug could emit `01:60`, which it did for Mecca. Both fixed upstream.

That last one had a test, and the test passed, because the format assertion used `/^\d{2}:\d{2}$/` and only ran against New York. That pattern matches `01:60` happily. It is now a strict `HH:MM` range check running across eight cities.

### Verification

62 integration tests, 12 of which boot a real Home Assistant and assert the exact entity IDs, names and states. Plus 25 new API cases across those eight cities. `hassfest`, `hacs/action`, the translations gates and the pytest job are all green on the [latest run](https://github.com/ummeco/praycalc-ha/actions).

Re-requesting review. Thanks again, this was a genuinely useful review and the integration is in much better shape for it.

PR #9256, opened 2026-07-16
Adds new integration [slettmayer/austria-public-transport]

slettmayer:
Adds the custom integration [`slettmayer/austria-public-transport`](https://github.com/slettmayer/austria-public-transport) — real-time Vienna (Wiener Linien) departures and Austrian rail (ÖBB) boards, trips, and service alerts.

## Checklist

## Links

Link to current release: https://github.com/slettmayer/austria-public-transport/releases/tag/v1.1.0
Link to successful HACS action (without the `ignore` key): https://github.com/slettmayer/austria-public-transport/actions/runs/33311635508/job/99257649908
Link to successful hassfest action (if integration): https://github.com/slettmayer/austria-public-transport/actions/runs/33311635508/job/99257649826


frenck (changes requested):
Hi @slettmayer, thanks for the submission! 👋

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

frenck (changes requested):
Hi @slettmayer, thanks for the submission, and for clearing the conflict! 👋

1. The `product_filter` option on `oebb_service_alerts` cannot do what the docs say it does. `__init__.py` lines 105 to 107 declare `vol.Optional("product_filter", default=1023): vol.All(int, vol.Range(min=1, max=1023))`, and line 191 passes the same default. The README, `docs/domain/OVERVIEW.md` and the 0.9.2 changelog entry all say the default is 65535 and that bit 12 (4096) selects private operators such as Westbahn and RegioJet, so anyone copying the documented call gets a validation error, and the private-operator alerts the changelog says were fixed cannot be requested at all. The 0.9.2 change landed only in `const.py` as `OEBB_ALL_PRODUCTS = 65535`; `__init__.py`, `services.yaml`, `strings.json` and both translation files still cap at 1023. Please raise the schema range and the selector to 65535 and default to `OEBB_ALL_PRODUCTS`, or roll the docs back to 1023 and retract the changelog line.

2. `coordinator.py` lines 36 to 40 hide every API failure. When `async_fetch_departures` returns an error dict, `_async_update_data` returns `self.data`, the previous poll, instead of raising `UpdateFailed`. `last_update_success` therefore stays true forever, entities never go unavailable, and a stop whose API has been down for hours keeps reporting a stale departure count that automations will act on as if it were current. Please `raise UpdateFailed(result["message"])` and let the coordinator own availability and backoff.

3. A missing countdown silently wipes out a stop. `api.py` line 61 always creates the `countdown` key, possibly as `None`, and line 68 sorts with `key=lambda x: x.get("countdown", 999)`. The default only applies when the key is absent, so a `None` value raises `TypeError: '<' not supported between instances of 'NoneType' and 'int'`. The blanket `except Exception` at line 84 catches it and returns `{"message": "No data"}`, so instead of an unsorted departure list the user loses every attribute for that stop with only a warning in the log. Please use `key=lambda x: x.get("countdown") if x.get("countdown") is not None else 999`.

4. `hacs.json` declares no `homeassistant` key, so HACS enforces no minimum core version. Nothing in the code needs a modern core, but the in-tree `custom_components/austria_public_transport/brand/` icons are only served from 2026.3.0, so on anything older the icons the 1.0.1 changelog promises are silently inert. Please declare the floor you actually want to support.

Two smaller things worth folding into the same release: the README has no screenshots, and `direct_only` ships on `oebb_trip_search` but is missing from the README's option table for that service.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

slettmayer:
Hi @frenck, all four points are addressed in [v1.1.0](https://github.com/slettmayer/austria-public-transport/releases/tag/v1.1.0), and the branch is rebased on the latest `master`.

**1. `product_filter`** — you were right that the 0.9.2 change only ever landed in `const.py`. Rather than roll the docs back, I raised the code to match them: `OEBB_ALL_PRODUCTS` (65535) is now the default and the range maximum in the service schema, the `services.yaml` selector, `strings.json` and both translations. Bit 12 (4096) for private operators is reachable, and the documented call no longer fails validation.

**2. Coordinator hiding failures** — `_async_update_data` now raises `UpdateFailed(result["message"])`; the stale-data fallback is gone and HA owns availability and backoff.

That change had a consequence worth flagging: `async_setup_entry` awaited `async_config_entry_first_refresh()` inside the stop loop, which was harmless while the coordinator never raised, but would have let one unreachable stop abort setup for *every* stop. Each stop is now refreshed with `async_refresh()`, and `ConfigEntryNotReady` is raised only when no stop succeeded — so a wrong RBL number leaves one unavailable entity beside working ones, while a genuine outage still retries the entry. The YAML platform path had the same loop and got the same treatment with `PlatformNotReady`.

**3. `None` countdown** — fixed with the key you suggested. Added a regression test that reproduces the exact failure (`TypeError` swallowed into `{"message": "No data"}`, losing every attribute for the stop) against the old sort key.

**4. `hacs.json` minimum version** — now declares `"homeassistant": "2026.3.0"`, the version from which the bundled `brand/` icons are actually served. Also stated in the README's install section.

Of the two smaller items: `direct_only` is now documented in the `oebb_trip_search` option table. **README screenshots are still missing** — I'd rather not hold the re-review on them, and they'll follow in a later release, but say the word if you'd prefer them in place first.

PR #81, opened 2019-12-16
Adds new plugin [Anonym-tsk/lovelace-starline-card]

Anonym-tsk:
Before you submit a pull request, please make sure you have the following:

- [x] You are submitting only 1 repository.
- [x] You repository is compliant with https://hacs.xyz/docs/publish/start
- [x] You have tested it with HACS by adding it as a custom repository

PR #9282, opened 2026-07-17
Adds new integration [FaserF/ha-speedport]

FaserF:
## Checklist

- [X] (For integrations only) I've added the [hassfest action](https://developers.home-assistant.io/blog/2020/04/16/hassfest/) to my repository.

## Links

Link to current release: https://github.com/FaserF/ha-speedport/releases/tag/v1.1.3
Link to successful HACS action (without the `ignore` key): https://github.com/FaserF/ha-speedport/actions/runs/29498746060/job/87622372471
Link to successful hassfest action (if integration): https://github.com/FaserF/ha-speedport/actions/runs/29498746060/job/87622372546

### DIFF
--- a/integration
+++ b/integration
@@ -966,6 +966,7 @@
   "FaserF/ha-kadermanager",
   "FaserF/ha-lidl",
   "FaserF/ha-openwrt",
+  "FaserF/ha-speedport",
   "FaserF/ha-traderepublic",
   "FaserF/ha-whatsapp",
   "Favio25/Kronoterm-homeassistant",


frenck (changes requested):
Hi @FaserF, thanks for the submission! 👋

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

FaserF:
Rebased on latest master, conflicts resolved.

frenck (changes requested):
Hi @FaserF, thanks for this one too, and for the rebase! 👋

1. The integration domain collides with a catalog entry that is already there. `pajew-ski/ha-speedport` has been in the `integration` catalog for a while, and both manifests declare `"domain": "speedport"`, both `hacs.json` files use the display name "Speedport", and both releases ship an asset called `speedport.zip`. Since both install into `custom_components/speedport/`, they are mutually exclusive on any one Home Assistant instance, the second install silently overwrites the first, and the HACS store would list two entries called Speedport with nothing to tell them apart. Your own issue #81 has a reporter saying they came from "die selbe Integration von wem anderes", so this is already biting people. Please pick a distinct domain (something like `speedport_faserf` or a name tied to the router family you target) and a distinct display name in `hacs.json`. Because the domain change also changes the catalog line, please open a fresh PR once that is out and I will close this one at that point.

2. Three parenthesis-less except clauses pin the integration to Python 3.14, but no floor is declared. `api.py` line 902 has `except ValueError, TypeError:`, and `sensor.py` lines 291 and 302 have the same shape. The syntax is fine; the problem is that `hacs.json` has no `homeassistant` key, and core pins `>=3.13.2` at 2026.2.0 and `>=3.14.2` at 2026.3.0, so everyone below HA 2026.3 gets a `SyntaxError` at import with no usable message. Please add `"homeassistant": "2026.3.0"`, or parenthesise the three tuples and declare the floor the rest of the code actually needs.

3. A dead or unauthenticated router looks like a healthy one. `api.py` lines 866 and 867 wrap the entire authenticated block, including `await self._ensure_auth()`, in `except Exception as exc: _LOGGER.debug(...)`. So a `SpeedportAuthError` or an unreachable router never reaches the coordinator; `_build_data` runs on an empty `raw` and returns a `SpeedportData` full of defaults, the refresh is marked successful, and entities publish `"Speedport"`, `""` and `0` as though they were live readings. That is worse than going unavailable, because it writes those values into history and any automation reading them. The same shape appears at lines 733 and 71. Please let the auth and connection errors propagate out of `get_all_data` and catch only per-endpoint failures.

4. Entities never go unavailable. `entity.py` lines 47 to 49 return `self.coordinator.data is not None`, and `binary_sensor.py`, `sensor.py` and `switch.py` each duplicate that override. Because the coordinator keeps `.data` from the last good poll, one successful refresh makes those entities permanently available: unplug the router and the public IP, DSL sync and WiFi switches keep showing the last known values forever. `image.py` line 82 already returns `self.coordinator.last_update_success`, which is the right answer. Please use that everywhere.

5. Device trackers and the update entity never write state after creation. `device_tracker.py` and `update.py` are both non-polling (`ScannerEntity` inherits `should_poll` false, and `update.py` line 43 sets it explicitly), and neither registers a coordinator listener, unlike `sensor.py`, `binary_sensor.py` and `switch.py` which all do. `device_tracker.py` line 53 registers `_add_new_devices`, which only ever adds entities. So a tracked client is frozen at whatever `is_connected` it had when first seen and never returns to `not_home`, which is the opposite of the README's "reliable presence detection for your home network". Please add the same `async_added_to_hass` hook calling `coordinator.async_add_listener(self.async_write_ha_state)` to both.

6. There is no reauthentication path. `coordinator.py` lines 64 to 70 catch `SpeedportAuthError`, retry the login once and then raise `UpdateFailed`, and the config flow has `async_step_user`, `async_step_reconfigure` and the options flow but no `async_step_reauth`, with no `reauth` block in `strings.json`. After a router password change the entry just logs update failures forever with no prompt. Your reconfigure step is nearly reusable for this.

7. The Wi-Fi passphrase reaches the log. `api.py` lines 465 to 467 log the first 300 characters of every raw response, including `data/WLANSettings.json`, whose payload carries `wlan_wpa_key`, the same field `image.py` reads to build the QR code. Line 410 logs the CSRF `httoken` the same way. Since `manifest.json` lists the integration under `loggers`, enabling debug is the first thing a user does before pasting a log into an issue. Your `diagnostics.py` redaction set already covers `wlan_key` and `wpa_key`, so this looks like an oversight rather than a choice. Please log `list(data.keys())` as you already do at lines 484 to 489, or redact before logging.

Two small ones: the README says the update interval defaults to 30 seconds while `const.py` line 22 sets `DEFAULT_UPDATE_INTERVAL = 60`; and `async_unload_entry` neither unregisters the five services nor closes the aiohttp session, so each reload leaks a session.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

PR #9294, opened 2026-07-17
Adds new integration [prusuino/ha_swiss_charging_stations]

prusuino:
## Checklist

## Links

Link to current release: https://github.com/prusuino/ha_swiss_charging_stations/releases/tag/v1.11.0
Link to successful HACS action (without the `ignore` key): https://github.com/prusuino/ha_swiss_charging_stations/actions/runs/33270256818/job/99147358447
Link to successful hassfest action (if integration): https://github.com/prusuino/ha_swiss_charging_stations/actions/runs/33270256818/job/99147358319

### DIFF
--- a/integration
+++ b/integration
@@ -2302,6 +2302,7 @@
   "Prosono/SMARTi_BaseComponent",
   "ProudElm/solaredgeoptimizers",
   "Proxy-alt/cync-lan",
+  "prusuino/ha_swiss_charging_stations",
   "ptimatth/GeorideHA",
   "ptrkoz/ha-smogtok",
   "PTST/LibreView-HomeAssistant",
--- a/integration
+++ b/integration
@@ -2396,6 +2396,7 @@
   "Prosono/SMARTi_BaseComponent",
   "ProudElm/solaredgeoptimizers",
   "Proxy-alt/cync-lan",
+  "prusuino/ha_swiss_charging_stations",
   "prusuino/ha_swiss_solar_reference_price",
   "ptimatth/GeorideHA",
   "ptrkoz/ha-smogtok",


frenck (changes requested):

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

prusuino:

prusuino:

prusuino:
Thanks for the review — this is addressed in **v1.9.0**, released just now.

- The old code also **deleted foreign Lovelace resource entries** that happened to match a
  previous card filename — that is gone too.
- Fixed: map markers a user un-hid by hand were re-hidden on every restart.
- The tariff endpoint stays plain HTTP because the SFOE GeoServer offers no TLS listener on
  any port; only public data is read and no credentials are sent. `const.py` and the README
  now say so explicitly rather than leaving it to be discovered.

frenck (changes requested):
Hi @prusuino, thanks for the submission, and for getting ahead of me on the dashboard question. 👋

I checked v1.9.0 and the Lovelace writing really is gone rather than just claimed gone: no dashboard module, no `LOVELACE_DATA`, nothing touching the storage collection, and the marker visibility now rides on `_attr_entity_registry_visible_default` with an explicit comment that an existing entity's visibility is never rewritten. That is exactly right, and the same fix is what your other two submissions need.

Two things left, and both bite specifically the multi-entry setup your own README recommends:

1. Map markers collide between two entries. `geo_location.py` builds `_UNIQUE_ID_PREFIX = f"{DOMAIN}_loc_"` and then `self._attr_unique_id = f"{_UNIQUE_ID_PREFIX}{location_id}"`, so the id comes from the site's ChargingStationId alone with nothing identifying the config entry. Setup step 3 tells the user to add the integration again for a different location or radius, and two overlapping radii (a 15 km home entry and a 15 km work entry in the same region) will both try to register a marker for every shared site. Home Assistant rejects the second with "does not generate unique IDs", so those sites silently vanish from the second entry's map with only a log line. Please include `entry.entry_id` in the id, and extend the one-time cleanup so existing markers are migrated rather than removed.

2. Markers never go unavailable. `ChargingSiteEvent` is a plain `GeolocationEvent` rather than a `CoordinatorEntity` and defines no `available`, so it stays true no matter what the coordinator did. When ich-tanke-strom.ch is unreachable your sensors correctly go unavailable, but every map marker keeps showing its last known "6/7 available" label indefinitely, which is the one moment a user is most likely to act on it and drive somewhere. It also undercuts the README's promise that entities stop updating rather than showing incorrect data. Please add an `available` property returning `self._coordinator.last_update_success`; the sync closure already has the coordinator to hand.

Three documentation items for the same release. README line 150 tells users to add the repository through HACS custom repositories, which stops being true once this merges. The entity table at line 23 documents `geo_location.ladestation_...`, but the object id comes from the localized entity name, so an English instance gets `geo_location.charging_station_...` and a template written from the table finds nothing. And `number.py` line 36 plus `select.py` lines 67, 95 and 135 hardcode absolute entity ids like `number.charging_stations_min_power_kw`; the unique ids are properly scoped so a second entry does not fail, but the registry resolves the clash by appending `_2`, so someone copying an automation between entries silently targets the wrong one. `sensor.py` already avoids this by putting the radius in the id, so the same treatment on those four would do it.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

prusuino:
Thanks for the second look — all five points are addressed in **v1.10.0**, released just now.

1. **Marker ids are entry-scoped.** The unique id is now `ich_tanke_strom_loc_<entry_id>_<site>`, so two overlapping radius entries register their own markers for a shared site. Existing registry entries are migrated in place on the first start (`async_update_entity(new_unique_id=...)`, only for the entry's own geo_location entries), so entity ids, history and user changes are kept. Verified on a live instance with 163 markers: 163 migrated, none duplicated, none lost.
2. **Markers go unavailable with the coordinator.** `ChargingSiteEvent` now exposes `available` returning `last_update_success`, and the sync listener writes state on every coordinator notification, so the flip reaches the state machine.
3. README installation describes the HACS default listing with the My Home Assistant link; the custom-repository path is down to one fallback sentence.
4. The entity table explains that marker object ids follow the localized name and shows the language-independent ways to address them (`geo_location_sources`, the `source` attribute in templates).
5. The four filter entities carry the radius in their suggested entity id, as the sensors already did; unique ids are unchanged, so existing entries keep their ids.

Also in this release: the minimum Home Assistant version is now 2025.4.0 — the strategy's marker labels use the map card's attribute label mode, which only exists since 2025.4.

prusuino:
Thanks for the review. Addressed in [v1.11.0](https://github.com/prusuino/ha_swiss_charging_stations/releases/tag/v1.11.0); the PR body now links that release and its action runs.

- Suggested entity ids carry the last four characters of the config entry id next to the radius or favorite name, so two entries with the same radius or two favorites with the same name no longer collide as `_2`. Existing entities keep their ids, unique ids unchanged.
- The kept registry row of a marker whose site leaves the radius is now explained in the code; the deselected per-plug-type sensors are removed from the registry.

Re-requesting review.

PR #9298, opened 2026-07-17
Adds new integration [MovingLlama/neoom]

MovingLlama:
## Checklist

## Links

Link to current release: <https://github.com/MovingLlama/neoom/releases/tag/v1.0.2>
Link to successful HACS action (without the `ignore` key): <https://github.com/MovingLlama/neoom/actions/runs/29603960811>
Link to successful hassfest action (if integration): <https://github.com/MovingLlama/neoom/actions/runs/29603960812>

### DIFF
--- a/integration
+++ b/integration
@@ -1963,6 +1963,7 @@
   "mossipcams/autosnooze",
   "Motorline/ha-mconnect",
   "mousehome-net/keba_one_ev",
+  "MovingLlama/neoom",
   "mpross512/ha-immich-frame",
   "mr-deamon/publibike_stations",
   "Mr-Groch/ambihue",


frenck (changes requested):
Hi @MovingLlama, thanks for the submission! 👋

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

frenck (changes requested):
Hi @MovingLlama, thanks for the submission, and for the rebase! 👋

**Please do this first, before anything else in this review.** Five scripts at the root of your repository hardcode live API keys: `query_cloud.py`, `query_cloud_sites.py` and `test_api.py` each carry a neoom CONNECT cloud token, and `query_beaam.py`, `query_site_state.py` and `test_api.py` each carry a BEAAM gateway key. Your site UUID and a LAN address are in there too. They are in the v1.0.4 release tarball and they are still on `main` as I write this. I have not tried them against anything, and I am deliberately not quoting them here, but you should assume they are compromised: anyone reading the repository can already see them.

What to do, in this order: rotate both keys in the neoom portal now, then `git rm --cached` those five scripts and commit. Note that having them in `.gitignore` does nothing, because they are already tracked. Rewriting history afterwards is worth doing, but the rotation is the part that actually protects you, so do that first and do not wait on the rest. This matters independently of whether this PR ever merges, which is why it is at the top.

Now the code items:

1. `hacs.json` declares `"homeassistant": "2024.1.0"`, but `sensor.py` returns `UnitOfPower.MEGA_WATT` at line 311, `UnitOfPower.GIGA_WATT` at line 313 and `UnitOfEnergy.GIGA_WATT_HOUR` at line 323. I checked those against core: all three are absent at 2024.10.0 and 2024.11.0 and first appear at 2024.12.0. On anything older, a gateway datapoint reporting MW, GW or GWh raises `AttributeError` inside `_map_unit` and that sensor never sets up. Please declare `"homeassistant": "2024.12.0"`.

2. A rejected key can never be corrected. `coordinator.py` raises `ConfigEntryAuthFailed` on a 401 at line 156, but the `except Exception as err: raise UpdateFailed(...) from err` at the end of `_ensure_config_loaded` catches its own `ConfigEntryAuthFailed` and converts it. Even where the exception does propagate correctly, `config_flow.py` defines no `async_step_reauth`, and core's `async_start_reauth_if_available` returns silently when the handler has none. So a rotated BEAAM key leaves the user with quietly unavailable entities and a generic update-failed line in the log, and the only way back is deleting and re-adding the entry. Please re-raise `ConfigEntryAuthFailed` from that handler and add a reauth step.

3. Controls can silently never appear. `switch.py` and `time.py` both return early when `local_coordinator.data` is empty and then require a non-empty settings map, and `_fetch_thing_settings` swallows every error and returns `None`. If the per-thing settings calls happen to fail or time out during the one first refresh, no switch, time, select-setting or number-setting entity is ever created: the user's battery, EV and heat-pump controls are simply absent, with nothing in the UI to explain it, until they manually reload the entry. Please track known things in the coordinator and add entities on later refreshes.

4. Responses are indexed without checking their shape. `coordinator.py` indexes `item["dataPointId"]` and `item["key"]` at lines 228 and 260, and reads `s["value"]` at line 266 after only checking `s.get("key")`; `number.py` line 231 calls `float(val)` with no guard. One partial payload from the gateway turns into a `KeyError` or `ValueError`, which in the coordinator fails the whole refresh and in `number.py` is raised inside a property, so it repeats every update. `NeoomSettingNumber.native_value` already does this correctly, so that pattern is right there to copy.

5. `custom_components/neoom/main.yml` is a GitHub Actions workflow for a different project: it rewrites `custom_components/innonet/manifest.json` and zips `innonet.zip`. Nothing breaks, but it gets copied into every user's `custom_components/neoom/` directory. Please delete it.

6. The README promises weather forecasts and statistical values from the cloud, but the cloud coordinator fetches only the site and the latest energy flow, and three cloud sensors are created, none forecast-related. It also documents a `beta` branch that does not exist, and the German text says the local poll is per-second while the code uses 15 seconds.

One more worth a look: `sensor.py` line 379 assigns `TOTAL_INCREASING` to anything whose key contains "ENERGY" outside a short unit blocklist, which catches non-cumulative values like stored battery energy. Every normal decrease then reads as a meter reset and inflates long-term statistics, which works against the Energy Dashboard support the README leads with.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

MovingLlama:
Hi @frenck,

Thank you very much for the detailed and thorough review! All points have been addressed in the codebase:

- **Credentials & Scripts:** Both the neoom CONNECT cloud token and the local BEAAM gateway API key have been rotated. All query and test scripts at the repository root (`query_cloud.py`, `query_cloud_sites.py`, `query_beaam.py`, `query_site_state.py`, `test_api.py`, `query_api.py`) have been removed from tracking via `git rm --cached` and added to `.gitignore`.
- **HACS Minimum Version:** Updated `hacs.json` to declare `"homeassistant": "2024.12.0"` so that `UnitOfPower.MEGA_WATT`, `UnitOfPower.GIGA_WATT`, and `UnitOfEnergy.GIGA_WATT_HOUR` are guaranteed to exist.
- **Authentication & Reauth Step:** 
  - `ConfigEntryAuthFailed` is now explicitly re-raised in `coordinator.py` upon HTTP 401 instead of being converted into an `UpdateFailed`.
  - Implemented `async_step_reauth` and `async_step_reauth_confirm` in `config_flow.py` (along with localized strings) to allow users to update credentials seamlessly.
- **Dynamic Entity Discovery:** Registered update listeners (`coordinator.async_add_listener`) across `switch`, `time`, `number`, `select`, and `sensor` platforms. If per-thing settings or data are not immediately available on the initial refresh, entities are discovered and added dynamically on subsequent refreshes without requiring a manual integration reload.
- **Payload Shape Guards:**
  - Hard dictionary indexing in `coordinator.py` has been replaced with safe `.get()` calls and validation of keys/values.
  - Guarded `float(val)` conversions in `number.py` with `try / except (ValueError, TypeError)`.
- **Workflow Cleanup:** Removed the leftover `custom_components/neoom/main.yml`.
- **README Corrections:**
  - Removed outdated promises of cloud weather forecasts/statistics (clarified that cloud is used for tariff data).
  - Replaced the non-existent `beta` branch references with HACS pre-release instructions.
  - Corrected the German polling interval description from per-second to 15 seconds.
- **Sensor State Class (`TOTAL_INCREASING`):** In `sensor.py`, non-cumulative energy datapoints (such as stored energy, battery reserves, capacities, and limits) are now mapped to `SensorStateClass.MEASUREMENT` rather than `TOTAL_INCREASING` to prevent false meter resets and long-term statistic inflation in the Energy Dashboard.

All changes have been released in **v1.0.8**. Ready for re-review!

PR #9342, opened 2026-07-19
Adds new integration [Idelmaeaen/HA-Universal-Modbus]

Idelmaeaen:
## Checklist

## Links

Link to current release: <https://github.com/Idelmaeaen/HA-Universal-Modbus/releases/tag/v1.1.1>
Link to successful HACS action (without the `ignore` key): <https://github.com/Idelmaeaen/HA-Universal-Modbus/actions/runs/33335727845>
Link to successful hassfest action (if integration): <https://github.com/Idelmaeaen/HA-Universal-Modbus/actions/runs/33335727797>

### DIFF
--- a/integration
+++ b/integration
@@ -1258,6 +1258,7 @@
   "Ibepower/Ibepower-Homeassistant-Integration",
   "IceSlam/ha-anycubic-cloud-connect",
   "ideaalab/gui-recorder",
+  "Idelmaeaen/HA-Universal-Modbus",
   "Identity-labs/ha-maison-protegee2",
   "ifMike/homeyHASS",
   "IgnacioHR/de-dietrich-c230-ha",
--- a/integration
+++ b/integration
@@ -1305,6 +1305,7 @@
   "Ibepower/Ibepower-Homeassistant-Integration",
   "IceSlam/ha-anycubic-cloud-connect",
   "ideaalab/gui-recorder",
+  "Idelmaeaen/HA-Universal-Modbus",
   "Identity-labs/ha-maison-protegee2",
   "ifMike/homeyHASS",
   "IgnacioHR/de-dietrich-c230-ha",


frenck (changes requested):
Hi @Idelmaeaen, thanks for the submission! 👋

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

frenck (changes requested):
Hi @Idelmaeaen, thanks for the submission, and for clearing the conflict! 👋

1. The "remove orphaned entities" button deletes the hub's own diagnostic sensors. `frontend.py` builds `current_unique_ids` purely from `_entry_profile(entry)["entities"]`, but the three diagnostic sensors register as `f"{entry.entry_id}_{description.key}"` with keys `last_response_time_ms`, `communication_error_count` and `last_successful_update`, which never appear in a profile. So every one of them always tests as stale and gets removed, even though the UI promises to remove only entities no longer present in the profile. Because this goes through `registry.async_remove`, it also purges their recorder history and long-term statistics (the response time is `MEASUREMENT`, the error count is `TOTAL_INCREASING`) and discards any rename, area or hidden setting the user had applied. Please union the diagnostic keys into `current_unique_ids`, something like `{f"{entry.entry_id}_{d.key}" for d in DIAGNOSTIC_SENSORS}`.

2. `hacs.json` declares no `homeassistant` key, but the code needs 2024.12.0. `config_flow.py` returns `UniversalModbusOptionsFlow()` with no arguments and reads `self.config_entry`, and that implicit property only exists on `OptionsFlow` from 2024.12.0; `frontend.py` also uses `StaticPathConfig` and `async_register_static_paths` (2024.7.0) and `__init__.py` uses `entry.runtime_data` (2024.6.0). Below 2024.12 HACS still offers the download and the user gets an `AttributeError` the moment they open Configure. Please add `"homeassistant": "2024.12.0"`. Worth saying explicitly: you read `self.config_entry` and never assign it, which is exactly right, so the setter removal in 2025.12 does not affect you.

3. The three diagnostic sensors are named in German only. `sensor.py` sets `name="Antwortzeit"`, `name="Kommunikationsfehler"` and `name="Letzte erfolgreiche Abfrage"` directly on the entity descriptions, bypassing `translations/` entirely, so an English installation gets `sensor.<device>_antwortzeit` and friends as both entity id and friendly name. That is all three diagnostic entities, and it sits oddly against the README's promise of English and German translations. Please give each a `translation_key` and add the names to `strings.json` and both translation files.

4. About 180 lines of the options flow are unreachable. `async_step_init` shows an empty-schema form and on submit goes straight to `async_create_entry`, so nothing ever returns a menu, which leaves `manage_entities`, `profile_info`, `add_entity`, `edit_entity`, `delete_entity`, `import_profile` and `export_profile` with no way in. Meanwhile `strings.json` advertises all four `init.menu_options`, so a reader or a translator expects an in-flow entity editor that does not exist. Please either return `async_show_menu` from `async_step_init` or delete the dead steps along with their strings.

5. `codec.py` truncates instead of rounding on the fast path: when the data type is not `float32` and scale is 1 with offset 0, it does `raw = int(value)`. v1.0.0 used `int(round(...))` here. A number entity with a 0.5 step set to 3.7 now writes 3 rather than 4. Please restore the rounding.

6. The README's Known Limitations still says "Strings, 64-bit values, bitfields, and custom encodings are not supported", but the feature list and the register table both document `int64` and `uint64` over four registers, which shipped in v1.0.2. Please drop the 64-bit part of that bullet.

One note rather than a finding: I went through `formula.py` carefully because of the `eval()`. The AST allowlist holds up. `Attribute`, `Subscript`, `Lambda`, comprehensions, f-strings and `Pow` are all absent from `_ALLOWED_NODES`, string constants are rejected, `Name` is restricted to the bound variable plus `scale`, `offset` and the seven functions, and `Call.func` must be one of those seven names, so the usual `__class__`/`__subclasses__` and `__import__` routes are all closed and I could not find a way out. Nice work on that. The one thing worth a line in the README is that a formula author is effectively an admin, since `coordinator.py` also sleeps for an unbounded admin-set `pulse_ms`.

Also worth updating: the PR body links v1.0.0 and its action runs, but v1.1.0 is what HACS will install.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

PR #9345, opened 2026-07-19
Adds new integration [mrtncode/ha-fritzbox-voicemail]

mrtncode:
## Checklist

## Links

Link to current release: <https://github.com/mrtncode/ha-fritzbox-voicemail/releases/tag/v1.0.4>
Link to successful HACS action (without the `ignore` key): <https://github.com/mrtncode/ha-fritzbox-voicemail/actions/runs/29680408334>
Link to successful hassfest action (if integration): <https://github.com/mrtncode/ha-fritzbox-voicemail/actions/runs/29680408334>


frenck (changes requested):
Hi @mrtncode, thanks for the submission! 👋

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

mrtncode:
@frenck, thank you for feedback! I have rebased the branch :)

frenck (changes requested):
Hi @mrtncode, thanks for the submission, and for clearing the conflict! 👋

Two things need to happen before this can go into the default store, and then a set of code items.

1. The dependency has to come from a package index, pinned. `manifest.json` requires `custom_fritzconnection@git+https://github.com/mrtncode/fritzconnection.git` with no tag, branch or commit. Every user's install therefore resolves to whatever happens to be on that fork's `master` at the moment they install, so two people installing a week apart can run different library code, and a force-push or an account compromise on one personal fork silently changes what executes inside every user's Home Assistant, with no release, no review and nothing visible in HACS. It also makes every install depend on `git` and on that repository staying public. Please publish `custom_fritzconnection` to PyPI and require a pinned version. Pinning the git URL to an immutable commit SHA would at least close the moving-target problem, but an index release is what this should be.

2. Issue #19 is still open and reproduced on the release you are submitting. Two independent users report the switch and the sensor never populating on v2.0.0, one of them (`thetechnikfreak`, 2026-08-22) on a fresh install a week after v2.0.1 shipped, and the only change between those two releases is the media_source title fix, so nothing in the submitted release addresses it. The reporter's log also shows the `via_device` warning from item 5 below verbatim. Those two entities are the whole integration, so please root-cause and close #19 and cut a release before this lands. I would rather wait than list something that does not come up for people.

3. The delete service silently retargets. `__init__.py` `_get_target_entries` ends with:

   ```python
   if not target_entries and (
       fallback := next(iter(hass.config_entries.async_entries(DOMAIN)), None)
   ):
       target_entries.add((fallback, 0))
   ```

   So if the target resolves to nothing (a renamed entity, a typo, an entity belonging to something else), the service does not fail: it substitutes the first config entry at TAM index 0. With `delete_mode: "all"` that means a mistargeted automation permanently deletes every message on an arbitrary answering machine, and voicemail recordings do not come back. Your "No active FritzBox Voicemail integration found" error is already there for exactly this case, but the fallback makes it unreachable. Please drop the fallback and let that error raise.

4. Playback and browsing always bind to the first config entry. Both `views.py` and `media_source.py` do `runtime_data = next(iter(self.hass.data[DOMAIN].values()))`, and the media-source identifier carries the TAM index but not the entry, so with two Fritz!Boxes the browser lists only the first box's messages and every playback is served by the first box's client. That undoes the multi-voicemail support v2.0.0 was built for. `self.hass.data[DOMAIN]` is also unguarded, and since the view is registered in `async_setup` and never removed, a request before any entry is added or after all are unloaded raises and returns a 500. Please put the entry id in the identifier and the URL, look it up explicitly, and 404 when it is missing.

5. `entity.py` makes each device its own parent. For TAM index 0 the suffix is empty, so `identifiers` is `(DOMAIN, entry_id)` and `via_device` is the identical tuple; for any other index the parent exists only if TAM 0 happened to be selected too. No hub device is ever created. At your declared floor this is a logged warning rather than an exception, but it is on core's removal path and it is the warning your users are pasting into #19. Please register one hub device for the Fritz!Box, give every TAM its own identifier including index 0, and point `via_device` at the hub.

6. No `ConfigEntryNotReady`, no `ConfigEntryAuthFailed`, no reauth. `__init__.py` builds the `FritzConnection` with no `try`/`except`, so a router that is rebooting at Home Assistant startup makes setup fail permanently with no retry. And `coordinator.py` wraps every exception into `UpdateFailed`, so a changed Fritz!Box password becomes an endless update-failure loop with no repair card, and `config_flow.py` has no `async_step_reauth` to offer anyway.

Three smaller ones: `config_flow.py` sets `_errors["base"] = "no_tams_found"`, which exists in neither `translations/en.json` nor `de.json`, so that user sees the raw key (meanwhile `connection` and `unknown` are translated but never set); the same file's bare `except Exception` turns a timeout, a wrong IP or a box with TR-064 disabled into "Username/Password is wrong", sending people after a credential problem they do not have; and `MailboxView` is registered at `__init__.py` line 124 in `async_setup` and again at line 158 per config entry, which just accumulates shadowed duplicate routes.

One process note: `.github/workflows/release.yml` force-pushes the release tag after publishing, so your tags are mutable and can move after they are cut. It works today, but it means the artifact HACS points at is not immutable, which is worth fixing while you are in there.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

mrtncode:
@frenck  Thank you for your detailled feedback!

1.) Good point. Honestly the fork is just fritzconnection with one more file and I didn't plan to change much there (except something breaks). So a pypi release is a little bit awkward for me. I pinned the version a the specific commit now (https://github.com/mrtncode/ha-fritzbox-voicemail/commit/6070621e7e182875128406c46629f81d2d650173). I hope that's okay. If not, I can publish on pypi :)

2.) + 5.) I think I fixed the bug and registered the Fritzbox as a Hub. I released a new version with the fix and closed the issue. (https://github.com/mrtncode/ha-fritzbox-voicemail/commit/ead28de211b8dce36af2b3fff0e8dc0f69d819b8) and (https://github.com/mrtncode/ha-fritzbox-voicemail/commit/0c1adcedcd07951eb0013b649f36012c71cbdc80)

3.) Good catch, I removed the fallback :) (https://github.com/mrtncode/ha-fritzbox-voicemail/commit/ead28de211b8dce36af2b3fff0e8dc0f69d819b8)

4.) Thank you for poiting this out. That's quit important 😁 Fixed that too (https://github.com/mrtncode/ha-fritzbox-voicemail/commit/fa5530b49b8d34689ca6da0aac7ede7832d40dca)

6.) I think I fixed it all :) (https://github.com/mrtncode/ha-fritzbox-voicemail/commit/ae21aca5483435b353d2369f2419392128053606 and https://github.com/mrtncode/ha-fritzbox-voicemail/commit/77842abbd3f48102c8f31aa73650b795246f475f)

I addressed the smaller suggestions (missing translations, better error handling in the config flow and double registered views) too. Thank you for poiting out.

Regarding the release action. I fixed that too. I no longer force push on the existing tag. Now I just create a .zip artifact with the right version.

I hope it addressed all things and it could be added to HACS default :)

PR #9355, opened 2026-07-19
Adds new integration [prusuino/ha_austrian_charging_stations]

prusuino:
## Checklist

## Links

Link to current release: https://github.com/prusuino/ha_austrian_charging_stations/releases/tag/v1.4.0
Link to successful HACS action (without the `ignore` key): https://github.com/prusuino/ha_austrian_charging_stations/actions/runs/33270254830/job/99147352977
Link to successful hassfest action (if integration): https://github.com/prusuino/ha_austrian_charging_stations/actions/runs/33270254830/job/99147352954

### DIFF
--- a/integration
+++ b/integration
@@ -2302,6 +2302,7 @@
   "Prosono/SMARTi_BaseComponent",
   "ProudElm/solaredgeoptimizers",
   "Proxy-alt/cync-lan",
+  "prusuino/ha_austrian_charging_stations",
   "ptimatth/GeorideHA",
   "ptrkoz/ha-smogtok",
   "PTST/LibreView-HomeAssistant",
--- a/integration
+++ b/integration
@@ -2396,6 +2396,7 @@
   "Prosono/SMARTi_BaseComponent",
   "ProudElm/solaredgeoptimizers",
   "Proxy-alt/cync-lan",
+  "prusuino/ha_austrian_charging_stations",
   "prusuino/ha_swiss_solar_reference_price",
   "ptimatth/GeorideHA",
   "ptrkoz/ha-smogtok",


frenck (changes requested):

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

prusuino:

prusuino:

prusuino:
Thanks for the review — this is addressed in **v1.2.0**, released just now.

- Fixed: map markers a user un-hid by hand were re-hidden on every restart.

frenck (changes requested):
Hi @prusuino, thanks for this one too, and for the three releases today. 👋

I checked v1.3.0 and the Lovelace writing is genuinely gone here as well: no dashboard collection, no `LOVELACE_DATA`, nothing touching storage, and the strategy replacement really does store nothing. The marker unique ids now carry the config entry with an in-place registry migration, and the markers go unavailable with the coordinator. That is all exactly right, and your changelog describes what shipped rather than overstating it, which I appreciate.

Three things left:

1. A revoked or rotated API key never asks for a new one. `coordinator.py` flattens everything into `UpdateFailed`:

   ```python
   except Exception as err:
       raise UpdateFailed(f"ladestellen.at unreachable: {err}") from err
   ```

   `_api_get` calls `resp.raise_for_status()`, so a 401 or 403 arrives as a `ClientResponseError` and lands in the same bucket as a genuine outage, and there is no `async_step_reauth` in the config flow. If E-Control revokes or rotates the key, every entity goes unavailable and the log repeats "ladestellen.at unreachable" every five minutes forever, with no re-authenticate card and nothing pointing at the key. Your config flow already knows the difference: `_is_auth_error` is right there. Please check it in the coordinator and raise `ConfigEntryAuthFailed`, then add an `async_step_reauth` and `async_step_reauth_confirm` pair.

2. The API key renders in cleartext. The schema uses a plain `str` for the key field, so it shows in the clear during setup and is prefilled visibly when reconfiguring. Please use a `TextSelector` with `type: password`, the way the other integrations in the store do.

3. One entity-id case is still open, narrower than the one you fixed. The filter entities now carry the radius, which does what your changelog says. But the config flow's unique id is `f"radius_{round(lat,2)}_{round(lon,2)}_{radius}"`, so two entries at *different locations with the same radius* are both allowed, and they then suggest the identical object id: a 15 km home entry and a 15 km work entry give `select.at_charging_stations_status_15km` and `..._15km_2`, with nothing to say which is which. The favorites have the same shape through `slugify(name)`, so two favorites at sites sharing a `station_name` collide too. Adding a short per-entry discriminator alongside the radius closes both.

One optional thing: when a site leaves the radius, `geo_location.py` calls `async_remove(force_remove=True)`, which drops the state but leaves the registry row behind as a "no longer provided" entity, so filter fiddling accumulates them over time. Your plug-type sensor path in `sensor.py` does clean the registry, so this looks unintentional, but keeping the row is also a defensible choice for a site that may come back into range. Either way is fine, it just deserves a comment saying which you meant.

The rest is in good shape: MIT licence at the root, version matching the tag, in-tree brand assets, four translation files with identical key counts and no shortfall, no domain collision (`rolandzeiner/ladestellen-austria` declares `ladestellen_austria`, you declare `ladestellen_at`), no secrets anywhere in the tree, no blocker-tier patterns in either language, and the bundled card escapes every freeform value through its `_escape` helper with the only unescaped interpolation being a lookup in a fixed colour map.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

prusuino:
Thanks for the review. Addressed in [v1.4.0](https://github.com/prusuino/ha_austrian_charging_stations/releases/tag/v1.4.0); the PR body now links that release and its action runs.

- 401/403 raise `ConfigEntryAuthFailed` in the coordinator; `async_step_reauth`/`reauth_confirm` added, `invalid_auth` reachable in the flow, and the generic exception handler no longer maps auth errors to `UpdateFailed`.
- API key entered through a masked `TextSelector` (password) in every dialog and no longer prefilled from an existing entry.
- Suggested entity ids carry the last four characters of the config entry id, so two entries with the same radius or two favorites with the same name no longer collide as `_2`. Existing entities keep their ids.
- The kept registry row of a marker whose site leaves the radius is now explained in the code; the deselected per-plug-type sensors are removed from the registry.

Re-requesting review.

PR #9356, opened 2026-07-19
Adds new integration [HomeBrainz/ha-inkVision]

11ado33:

## Links

Link to current release: https://github.com/HomeBrainz/ha-inkVision/releases/tag/v1.0.12
Link to successful HACS action (without the `ignore` key): https://github.com/HomeBrainz/ha-inkVision/actions/runs/29690447893/job/88202525388
Link to successful hassfest action (if integration): https://github.com/HomeBrainz/ha-inkVision/actions/runs/29690447893/job/88202525900

### DIFF
--- a/integration
+++ b/integration
@@ -1219,6 +1219,7 @@
   "home-assistant-HomeWhiz/home-assistant-HomeWhiz",
   "Home-Is-Where-You-Hang-Your-Hack/sensor.goveetemp_bt_hci",
   "HomeAssistant-Mods/home-assistant-miele",
+  "HomeBrainz/ha-inkVision",
   "Homie-Assistance/norwegian-parcel-tracker",
   "hopkins-tk/home-assistant-aseko-local",
   "hostcc/hass-gs-alarm",


frenck (changes requested):
Hi @11ado33, thanks for the submission! 👋

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

11ado33:
Done.

frenck (changes requested):
Hi @11ado33, thanks for the submission, and for the rebase! 👋

1. The integration may not load at all. `import async_timeout` appears in `__init__.py`, `config_flow.py`, `switch.py` and `button.py`, but `manifest.json` requires only `aiohttp>=3.8.0` and `websockets>=11.0`. I checked core: `async-timeout` is in `homeassistant/package_constraints.txt` as an upper-bound pin but not in `requirements.txt`, so it is constrained if something else installs it and never installed on its own. Core's own `pyproject.toml` also bans the import outright with the message "use asyncio.timeout instead". On an installation where nothing else has pulled it in, setup dies with `ModuleNotFoundError`. Please switch all four to the stdlib `asyncio.timeout`.

2. With two devices, every service call goes to the wrong one. `_get_coordinator_for_device` looks the device up and then does:

   ```python
   for entry_id, coordinator in hass.data[DOMAIN].items():
       if device.identifiers:
           for identifier in device.identifiers:
               if identifier[0] == DOMAIN:
                   return coordinator
   ```

   It only checks that the identifier's namespace is your domain and never compares the value, and `entry_id` is unused. So `set_brightness`, `display_text`, `restart_device` and `set_timezone` all land on whichever coordinator the dict happens to yield first, and a user with two inkVision clocks restarts or retitles the wrong one. Please match `identifier[1]` against the coordinator's host or MAC, or resolve through `device.config_entries` and index `hass.data[DOMAIN]` by that entry id.

3. HTTPS media URLs are silently downgraded to plaintext. `media_player.py` `_resolve_media_id` ends with:

   ```python
   if resolved.startswith("https://"):
       parsed = urlparse(resolved)
       resolved = urlunparse(parsed._replace(scheme="http"))
   ```

   I understand the Atmos firmware only takes plain HTTP, but this runs after `async_process_play_media_url`, so it applies to any URL the user passes, including a remote HTTPS radio stream, and including Home Assistant's own signed media-source URLs whose `authSig` token then travels in the clear. Please restrict the downgrade to hosts on the local network, or put it behind an explicit option, and log at warning level when it happens rather than debug.

4. The integration ships with no icon on modern Home Assistant. There is no `custom_components/homebrainz/brand/` directory; the images live at `icon.png`, `icons/logo.svg` and `www/icon.png`, and `brands_submission/submit.sh` targets `home-assistant/brands`, which no longer accepts custom integrations. In-tree branding is the only path now: core reads `Path(integration.file_path) / "brand"` and gates on `"brand"` being a top-level file. Please move the images into `custom_components/homebrainz/brand/` as `icon.png`, `icon@2x.png` and `logo.png`, and drop `brands_submission/`.

5. `hacs.json` has two problems. It sets `"filename": "homebrainz.zip"` while `zip_release` is unset and release v1.0.13 carries no assets, so the key is dead today and would break installs the moment `zip_release` were added. And there is no `homeassistant` key, although `__init__.py` calls `async_update_entry(entry, version=2)` and `version=3`; that keyword arrived in core 2024.3.0 and is absent at 2024.2.0, so on anything older the migration raises `TypeError` and the entry is stuck. Please drop `filename` and add `"homeassistant": "2024.3.0"`.

6. `manifest.json` says `"version": "1.0.12"` while the tag is `v1.0.13`, so HACS shows users a version one release behind what they installed. The root `VERSION` file says `0.0.10`, which is a third answer.

7. `set_timezone` is registered in `__init__.py` but `services.yaml` describes only four services, so it has no name, description or device selector in Developer Tools and cannot be built in the visual editor, even though the README documents it.

Three smaller ones: `sensor.py` dereferences `self.coordinator.data` at line 220 without the guard used on the line above, and does `sensors["bme680"].get(...)` after only an `in` test in four places, so firmware returning a null or a list raises inside a state property, while your own `get_sensor_value` helper does check the type; the WebSocket loop `break`s permanently after five failures, so a device reboot or an access-point roam past the backoff window drops a `local_push` integration to five-minute polling until the entry is reloaded; and `_LOGGER.info` writes the whole sensor payload on every push message plus user display text, which fills the log at the device's push rate (nothing sensitive, just volume).

Finally, please refresh the PR body: it links release v1.0.12 and its CI run, while v1.0.13 is current. The validation workflow is green on the current commit, so it is just the links.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

PR #9362, opened 2026-07-19
Adds new integration [prusuino/ha_swiss_transport]

prusuino:
## Checklist

## Links

Link to current release: https://github.com/prusuino/ha_swiss_transport/releases/tag/v1.4.0
Link to successful HACS action (without the `ignore` key): https://github.com/prusuino/ha_swiss_transport/actions/runs/33270259104/job/99147364374
Link to successful hassfest action (if integration): https://github.com/prusuino/ha_swiss_transport/actions/runs/33270259104/job/99147364215

### DIFF
--- a/integration
+++ b/integration
@@ -2302,6 +2302,7 @@
   "Prosono/SMARTi_BaseComponent",
   "ProudElm/solaredgeoptimizers",
   "Proxy-alt/cync-lan",
+  "prusuino/ha_swiss_transport",
   "ptimatth/GeorideHA",
   "ptrkoz/ha-smogtok",
   "PTST/LibreView-HomeAssistant",
--- a/integration
+++ b/integration
@@ -2397,6 +2397,7 @@
   "ProudElm/solaredgeoptimizers",
   "Proxy-alt/cync-lan",
   "prusuino/ha_swiss_solar_reference_price",
+  "prusuino/ha_swiss_transport",
   "ptimatth/GeorideHA",
   "ptrkoz/ha-smogtok",
   "PTST/LibreView-HomeAssistant",


frenck (changes requested):

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

prusuino:

prusuino:

prusuino:
Thanks for the review — this is addressed in **v1.3.0**, released just now.

frenck (changes requested):
Hi @prusuino, thanks for this one too. 👋

Same as your other two: I checked v1.3.1 and the Lovelace writing really is gone. No dashboard collection, no storage access, and the only `hass.callWS` calls in the card are the two read-only registry lists. Your 1.3.0 changelog describes that change honestly instead of glossing it, which I appreciate.

Three things:

1. A bad OJP token fails silently forever. `coordinator.py` wraps the enrichment call in `except Exception: _LOGGER.debug("OJP real-time enrichment failed", exc_info=True)`, and `async_fetch_ojp_stationboard` raises through `raise_for_status()` on a 401 or 403. The config flow never probes the token either: it just strips and stores whatever was pasted. So somebody who mistypes their token, or whose token expires, sees no error in the setup form and nothing in the log at default level, and simply never gets the cancellations, occupancy and disruption banners the README promises. The board looks exactly like the no-token case. Keeping enrichment best-effort at runtime is the right call, but being silent about *why* is not: please log the first auth failure once at warning with the status code, and validate the token in the flow so a bad one comes back as `invalid_auth`.

2. `SECURITY.md` is out of date and in the one place it matters most. It says the integration "only reads public timetable data from transport.opendata.ch over HTTPS, it needs no credentials, handles no personal data, and has no write access to any external system". Since 1.1.0 it optionally takes an OJP bearer token, and it reverse-geocodes station coordinates through Nominatim. Your README and `NOTICE.md` both describe those accurately, so this is staleness rather than a false claim, but a security reporter reading that page is told there is no credential surface when there is one. One sentence fixes it.

3. Same entity-id shape as your Austrian one. `sensor.py` sets `self.entity_id = f"sensor.swiss_transport_{slugify(name)}"` from the station name rather than the entry, so two boards for the same station (Zürich HB filtered to trains, and another to trams) both want the same object id and the second silently becomes `..._2`. The unique id is correctly scoped to the entry so nothing is dropped, but the README says the object id follows the stop name, and for the second board it quietly does not.

Two smaller ones. `frontend.py` sets its registered flag before the `await` — good for the double-registration race — but never clears it if `async_register_static_paths` raises, and `__init__.py` catches and continues, so one transient failure leaves the card unserved for the rest of that Home Assistant run with no retry. And the options flow rebuilds `new` from scratch, so saving a connection entry's result count drops the token `_apply_token_everywhere` had synced into it; nothing user-visible today since connection coordinators never use OJP and `_resolve_ojp_token` scans all entries, but the single-source-of-truth invariant your docstring asserts is not actually held. Seeding from `dict(self.config_entry.options)` covers it.

One thing I looked at and decided against raising as a finding, in case it is useful: `coordinator.py` uses stdlib `ET.fromstring` on the OJP response. There is no XXE exposure there, and hardening it would mean adding `defusedxml` as a requirement since it is not in core, so leaving it is reasonable for an authenticated government endpoint.

The rest is in good shape: MIT licence at root, version matching the tag, in-tree brand assets, four translation files at exact 57-key parity, no name or domain collision (core's is `swiss_public_transport`, yours is `swiss_transport`), no secrets in the tree, nothing in either blocker tier, and the card's `_escape` helper is applied to every vendor-supplied value that reaches `innerHTML`, with the unescaped ones all internal constants or number-gated.

Also worth refreshing: the PR body still links v1.0.0 and its CI run, three minor versions behind what I reviewed.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

prusuino:
Thanks for the review. Addressed in [v1.4.0](https://github.com/prusuino/ha_swiss_transport/releases/tag/v1.4.0); the PR body now links that release and its action runs.

- The real-time token is validated with a probe request when entered; a rejected token is shown in the form, and a 401/403 during enrichment is logged once at WARNING (not repeated until the token was accepted again).
- Options flow seeds from the existing options; saving a connection entry no longer drops the token copy.
- Suggested entity ids carry the last four characters of the config entry id, so two stops that slugify to the same text no longer collide. Existing entities keep their ids.
- SECURITY.md describes the optional opentransportdata.swiss token and the one-time Nominatim lookup.

Re-requesting review.

PR #9377, opened 2026-07-20
Adds new integration [luanche/llm_smart_assistant]

luanche:
## Checklist

## Links

Link to current release: <https://github.com/luanche/llm_smart_assistant/releases/tag/v1.1.2>
Link to successful HACS action (without the `ignore` key): <https://github.com/luanche/llm_smart_assistant/actions/runs/29753303319/job/88389139405>
Link to successful hassfest action (if integration): <https://github.com/luanche/llm_smart_assistant/actions/runs/29753303319/job/88389139457>


frenck (changes requested):
Hi @luanche, thanks for the submission! 👋

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

frenck (changes requested):
Hi @luanche, thanks for the submission. 👋

I have to close this one, and I want to be clear about why, because the reason is serious and it is fixable.

**The integration serves a Home Assistant long-lived access token to unauthenticated callers.** In `custom_components/llm_smart_assistant/__init__.py`, `ChatPanelView` is registered at `/api/llm_smart_assistant/chat_panel` with `requires_auth = False`, and its `get` handler does:

```python
access_token = ""
for coord in hass.data.get(DOMAIN, {}).values():
    if hasattr(coord, 'access_token') and coord.access_token:
        access_token = coord.access_token
        break
if access_token:
    script = f'<script>window.CONFIGURED_ACCESS_TOKEN={json.dumps(access_token)};</script>'
    current_html = current_html.replace("</head>", script + "</head>")
```

Your own `translations/en.json` describes that field as "Long-lived access token for the AI Chat panel. Generate one in your HA profile". So for any user who has filled it in, anybody who can reach their Home Assistant HTTP port, with no credentials whatsoever, can request that URL and read a token that grants full API control of the instance: every light, lock, camera and automation. On a port-forwarded or reverse-proxied instance that is the open internet. The `require_admin=True` you set on the sidebar panel does not help, because this is a separate route that never consults it.

Three more views share the same root cause. `/api/llm_smart_assistant/suggestions` is unauthenticated and calls the LLM, so an anonymous requester can spend the user's API credits at will. `/api/llm_smart_assistant/history` returns seven days of merged conversation history to anonymous callers. `/api/llm_smart_assistant/chat_js` is unauthenticated too.

The fix is small: delete `requires_auth = False` from all four views. The default is `True`, and your panel already sends `Authorization: Bearer` on every call, with a token-discovery chain in `index.html` that covers the iframe case, so authentication should just work. Then stop injecting the token into the HTML entirely. While you are there, the `?auth_token=` query string in `panel/chat.js` and the `postMessage(..., '*')` wildcard origin next to it should both go, since a token in a URL lands in logs and history.

I am closing rather than leaving this open because a repository in the default store is installed by people who will never read its source, and this needs a released fix rather than a promise. Once a release is out with those four views authenticated, please open a fresh PR through the publishing wizard and I will pick it straight up.

Five other things worth fixing in the same release:

1. `hacs.json` declares `"homeassistant": "2024.6.0"`, but your options flow reads the implicit `self.config_entry` property, which does not exist on `OptionsFlow` until 2024.12.0. On 2024.6 through 2024.11 opening Configure raises `AttributeError`. Please declare `2024.12.0`. Do not instead assign `self.config_entry` in the constructor: that setter was removed in 2025.12.0 and would break current versions.

2. `services.py` `_validate_service_call` checks the entity whitelist with `if allowed_entities and target:`, so a step like `{"action":"call_service","domain":"light","service":"turn_off"}` with no target skips the whitelist entirely and Home Assistant applies it to every entity in the domain. Someone who whitelisted one lamp gets the whole house turned off. Please reject a call whose resolved target is empty when a whitelist is set.

3. `async_unload_entry` never calls `async_unload_platforms`, so reloading leaves the three sensors registered against a dead coordinator and a reload adds a second set. The per-entry services are not removed on unload either.

4. `panel/index.html` puts LLM-controlled values into `innerHTML` unescaped in three places: the `expression` at the `trigHtml` construction, and the `weekdays` and `days_of_month` values. Your `escapeHtml` helper is right there and used on neighbouring fields. Since those values are persisted verbatim by `async_create_automation`, a prompt injection reaching the model through a device name or a description can execute script in the admin panel where a token is in scope. Escape all three, and validate those fields at creation time rather than only at fire time.

5. `chat`, `create_automation`, `update_automation` and `get_automations` are registered per config entry in closures over one coordinator, so with two instances the second setup overwrites the first and all four always act on the last one loaded. Your README advertises multi-instance support. `process_input` and `toggle_automation` already do this correctly with an `entry_id` filter.

For what it is worth, plenty here is done well: MIT licence at the root, manifest version matching the tag, in-tree brand assets, English and Simplified Chinese at full 61-key parity, no committed secrets anywhere in the tree, no obfuscation, nothing sent to any endpoint you do not own, and a hand-rolled recursive-descent parser for boolean trigger expressions instead of reaching for `eval`, which was the right instinct. The authentication on those four views is what stands between this and the catalog.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

luanche:
@frenck thx for your feedback, I will try to fix them and rise another PR

PR #9393, opened 2026-07-21
Adds new integration [FaserF/ha-finanzfluss]

FaserF:
## Checklist

- [X] (For integrations only) I've added the [hassfest action](https://developers.home-assistant.io/blog/2020/04/16/hassfest/) to my repository.

## Links

Link to current release: https://github.com/FaserF/ha-finanzfluss/releases/tag/v1.0.0
Link to successful HACS action (without the `ignore` key): https://github.com/FaserF/ha-finanzfluss/actions/runs/29833835316/job/88645437600
Link to successful hassfest action (if integration): https://github.com/FaserF/ha-finanzfluss/actions/runs/29833835316/job/88645437599


frenck (changes requested):
Hi @FaserF, thanks for the submission! 👋

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

FaserF:
Rebased on latest master, conflicts resolved.

frenck (changes requested):
Hi @FaserF, thanks for coming back to this and for clearing the conflict.

The rebase is resolved, so I did the full review pass. I reviewed against `v1.0.2` (published 2026-08-25), which is the current release; the PR body still links `v1.0.0`, worth updating.

I cannot list this yet. There is one hard blocker, plus a set of items I would like addressed in the same round.

## Blocker: the released zip does not parse on the Python that Home Assistant runs

`custom_components/finanzfluss/config_flow.py` line 284, in the `finanzfluss.zip` asset attached to `v1.0.2`:

```python
except CannotConnectError, FinanzflussAPIError:
```

Unparenthesized multiple exception types is PEP 758, new in Python 3.14. On Python 3.13 and earlier it is a `SyntaxError`, not a warning. I extracted the published asset and compiled it: `SyntaxError: multiple exception types must be parenthesized`, line 284.

Home Assistant's `REQUIRED_PYTHON_VER` is `(3, 13, 2)` through the 2026.2 line and only becomes `(3, 14, 2)` in 2026.3.0. `manifest.json` sets `"config_flow": true`, so Home Assistant imports this module the moment a user tries to add the integration. On any Home Assistant release up to and including 2026.2.x, that import raises and the integration cannot be configured at all.

The fix is one line:

```python
except (CannotConnectError, FinanzflussAPIError):
```

This is still present on `main` today. Three commits in the `v1.0.2` notes are aimed at exactly this class of problem ("Parenthesize exception types in all repos", "ensure parenthesized exceptions", "Fix exception syntax") and all three missed this file. CI does not catch it because the ruff target, the mypy `python_version` and the workflow's installed interpreter are all pinned to 3.14, which is the single Python version where the bug is invisible. Worth adding a compile check on the oldest Python you intend to support.

While you are in that file: line 42 annotates a return type of `FinanzflussOptionsFlow`, which is defined later at line 152, and the file has no `from __future__ import annotations`. That also only works on 3.14. Same fix either way.

## Blocker: `hacs.json` declares no version floor

```json
{
  "name": "Finanzfluss Copilot",
  "render_readme": true,
  "zip_release": true,
  "filename": "finanzfluss.zip",
  "country": ["DE"]
}
```

There is no `homeassistant` key, so HACS offers this to every Home Assistant version, including all the ones where the syntax above is fatal. These two are one problem: either parenthesize and declare the floor the APIs actually need (`ConfigFlowResult` needs 2024.4.0, the `config_entry=` keyword on `DataUpdateCoordinator` needs 2024.11.0, so `"homeassistant": "2024.11.0"`), or keep the 3.14 syntax and declare `"homeassistant": "2026.3.0"`. The first is much better for your users.

## The reauth flow is raised but not implemented

`coordinator.py` line 140 raises `ConfigEntryAuthFailed` when the token refresh fails. That tells Home Assistant to start a reauth flow, but `FinanzflussConfigFlow` has no `async_step_reauth` or `async_step_reauth_confirm`, and `strings.json` has no `config.step.reauth` entries. The only reauth-named step in the tree is `async_step_reauth_mfa` on the options flow, which is a different flow entirely. So a user whose credentials expire gets a repair notification that leads nowhere, which is why the README ends up telling them to go through Options instead. Either implement the config flow reauth steps or stop raising `ConfigEntryAuthFailed`.

## Stale financial data is served as if it were current

`coordinator.py` lines 70 to 77:

```python
if self._backoff_until and now < self._backoff_until:
    LOGGER.warning("In backoff period until %s ...", self._backoff_until)
    if self.data:
        LOGGER.info("Returning cached coordinator data during backoff.")
        return self.data
```

Returning normally from `_async_update_data` sets `last_update_success` to true, so every entity stays available and the recorder keeps writing the cached values as fresh observations. `_handle_failure` sets `_backoff_until` up to 24 hours ahead on a 403 or 429, and up to 4 hours otherwise. For that whole window the user sees a net worth and account balances that look live and are not, with nothing but an info log line to say so.

Backing off is the right instinct; replaying the cache is not the way to do it. Raise `UpdateFailed` during backoff and let the entities go unavailable, which is the honest signal and the one automations can act on.

## The integration tells users their password is encrypted, and it is not

README line 453 says "Credentials stored **encrypted** in the HA config entry store", and the setup dialog itself says "Your password will be encrypted." in both `strings.json` and `translations/en.json`.

`config_flow.py` line 85 writes `CONF_PASSWORD` straight into `entry.data`, and Home Assistant persists config entries as plaintext JSON in `.storage/core.config_entries`. There is no at-rest encryption. For a personal finance integration this is a promise users will act on, so please remove both claims or replace them with an accurate description.

## Diagnostics disclose the account email

`TO_REDACT` in `diagnostics.py` covers the password, both tokens, the refresh token, the uuid and the iban, but not `email`, and `config_flow.py` line 85 puts `CONF_EMAIL` into `entry.data`. The entry title is also set to the email at line 83, and diagnostics include the title, so it appears twice. Add `email` to `TO_REDACT` and use something other than the raw address as the entry title.

## Five unique IDs are not scoped to the config entry

Thirteen of the eighteen sensors interpolate `entry_id` into `_attr_unique_id`. These five use only an API-supplied identifier:

- `sensor.py:122` `finanzfluss_account_{account_id}`
- `sensor.py:500` `finanzfluss_budget_bucket_{bucket_id}`
- `sensor.py:553` `finanzfluss_budget_bucket_remaining_{bucket_id}`
- `sensor.py:818` `finanzfluss_investment_position_{position_id}`
- `sensor.py:872` `finanzfluss_exemption_order_{order_id}`

The config flow keys uniqueness on the lowercased email, so two accounts in one household are allowed. If those ids are per-user rather than globally unique, the second entry's entities collide with the first's and are dropped by the entity registry without an error the user will see. Scoping all eighteen to `entry_id` costs nothing and removes the question.

## Smaller items

- `CONF_FALLBACK_CALCULATION` is a real options toggle with a translation string, but the README options table lists only the update interval, new email and new password.
- README line 70 tells manual installers to extract a `custom_components/finanzfluss/` folder, but the release zip has the files at its root, so following it produces the wrong layout.
- `manifest.json` at tag `v1.0.2` says `"version": "1.0.2-dev0"`. The workflow rewrites it correctly inside the zip, so installs are fine, but anyone installing from source gets the dev string.
- `const.py:16` defines `API_CASHFLOW_HISTORY`, which nothing uses.
- The four PNGs under `brand/` are 165 KB of the 180 KB payload. Home Assistant serves brand assets from the brands CDN, not from the integration directory, so they are dead weight in every install.

Push a release with at least the syntax fix and the version floor and I will take another look.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

PR #9278, opened 2026-07-17
Adds new integration [prusuino/ha_swiss_electricity_price]

prusuino:
## Checklist

- [x] I've read the [publishing documentation](https://hacs.xyz/docs/publish/start).
- [x] I've added the [HACS action](https://hacs.xyz/docs/publish/action) to my repository.
- [x] (For integrations only) I've added the [hassfest action](https://developers.home-assistant.io/blog/2020/04/16/hassfest/) to my repository.
- [x] The actions are passing without any disabled checks in my repository.
- [x] I've added a link to the action run on my repository below in the links section.
- [x] I've created a new release of the repository after the validation actions were run successfully.

## Links

Link to current release: https://github.com/prusuino/ha_swiss_electricity_price/releases/tag/v1.4.0
Link to successful HACS action (without the `ignore` key): https://github.com/prusuino/ha_swiss_electricity_price/actions/runs/33271157301/job/99149751735
Link to successful hassfest action (if integration): https://github.com/prusuino/ha_swiss_electricity_price/actions/runs/33271157301/job/99149751657

PR #9251, opened 2026-07-16
Adds new integration [prusuino/ha_swiss_earthquakes]

prusuino:
## Checklist

- [x] I've read the [publishing documentation](https://hacs.xyz/docs/publish/start).
- [x] I've added the [HACS action](https://hacs.xyz/docs/publish/action) to my repository.
- [x] (For integrations only) I've added the [hassfest action](https://developers.home-assistant.io/blog/2020/04/16/hassfest/) to my repository.
- [x] The actions are passing without any disabled checks in my repository.
- [x] I've added a link to the action run on my repository below in the links section.
- [x] I've created a new release of the repository after the validation actions were run successfully.

## Links

Link to current release: https://github.com/prusuino/ha_swiss_earthquakes/releases/tag/v1.5.0
Link to successful HACS action (without the `ignore` key): https://github.com/prusuino/ha_swiss_earthquakes/actions/runs/33271163075/job/99149766380
Link to successful hassfest action (if integration): https://github.com/prusuino/ha_swiss_earthquakes/actions/runs/33271163075/job/99149766232

PR #9249, opened 2026-07-16
Adds new integration [prusuino/ha_swiss_avalanche_bulletin]

prusuino:
## Checklist

- [x] I've read the [publishing documentation](https://hacs.xyz/docs/publish/start).
- [x] I've added the [HACS action](https://hacs.xyz/docs/publish/action) to my repository.
- [x] (For integrations only) I've added the [hassfest action](https://developers.home-assistant.io/blog/2020/04/16/hassfest/) to my repository.
- [x] The actions are passing without any disabled checks in my repository.
- [x] I've added a link to the action run on my repository below in the links section.
- [x] I've created a new release of the repository after the validation actions were run successfully.

## Links

Link to current release: https://github.com/prusuino/ha_swiss_avalanche_bulletin/releases/tag/v1.3.0
Link to successful HACS action (without the `ignore` key): https://github.com/prusuino/ha_swiss_avalanche_bulletin/actions/runs/33271161722/job/99149763042
Link to successful hassfest action (if integration): https://github.com/prusuino/ha_swiss_avalanche_bulletin/actions/runs/33271161722/job/99149763205

PR #9407, opened 2026-07-21
Adds new integration [prusuino/ha_innoxel_master3]

prusuino:
## Checklist

## Links

Link to current release: https://github.com/prusuino/ha_innoxel_master3/releases/tag/v1.6.0
Link to successful HACS action (without the `ignore` key): https://github.com/prusuino/ha_innoxel_master3/actions/runs/33270253280/job/99147349138
Link to successful hassfest action (if integration): https://github.com/prusuino/ha_innoxel_master3/actions/runs/33270253280/job/99147349071

### DIFF
--- a/integration
+++ b/integration
@@ -2302,6 +2302,7 @@
   "Prosono/SMARTi_BaseComponent",
   "ProudElm/solaredgeoptimizers",
   "Proxy-alt/cync-lan",
+  "prusuino/ha_innoxel_master3",
   "ptimatth/GeorideHA",
   "ptrkoz/ha-smogtok",
   "PTST/LibreView-HomeAssistant",
--- a/integration
+++ b/integration
@@ -2396,6 +2396,7 @@
   "Prosono/SMARTi_BaseComponent",
   "ProudElm/solaredgeoptimizers",
   "Proxy-alt/cync-lan",
+  "prusuino/ha_innoxel_master3",
   "prusuino/ha_swiss_solar_reference_price",
   "ptimatth/GeorideHA",
   "ptrkoz/ha-smogtok",


frenck (changes requested):

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

prusuino:

prusuino:

prusuino:
Heads-up: after your reviews on #9249, #9251 and #9278 I checked this repository for the same issues. Only one applied, fixed in v1.5.3:

- `hacs.json` now declares `"homeassistant": "2024.12.0"` instead of 2024.1.0. The options flow relies on `OptionsFlow.config_entry`, which older versions do not provide, so the dialog would have failed there. **No functional changes** — this release only corrects the declared minimum version.

This integration has no dashboard code and already shipped `translations/en.json`. The SOAP endpoint is a local Innoxel controller that offers no TLS; `SECURITY.md` documents that.

Release: https://github.com/prusuino/ha_innoxel_master3/releases/tag/v1.5.3 — validation is green.

frenck (changes requested):
Hi @prusuino, thanks for rebasing, and for going back over your own repos against the findings from the other submissions before I got here. That saved a round on the dashboard question: this repository ships no frontend code at all, so nothing from that thread applies.

I reviewed against `v1.5.3`. The PR body still links `v1.2.1`, eleven releases back; please point it at the current one.

## Blocker: the declared floor is three minor releases too low

`config_flow.py` line 8:

```python
from homeassistant.helpers.service_info.ssdp import SsdpServiceInfo
```

`homeassistant/helpers/service_info/ssdp.py` does not exist in Home Assistant 2025.1.x or anything earlier; it first appears in 2025.2.0. I checked the file directly at both tags. `hacs.json` declares `"homeassistant": "2024.12.0"`.

`manifest.json` also declares an `ssdp` discovery block, so Home Assistant loads this module to match discoveries rather than only when someone opens the add dialog. On 2024.12.x through 2025.1.x the import raises and the integration cannot be configured or discovered at all.

The 2024.12.0 reasoning in the `v1.5.3` notes is right as far as it goes; `OptionsFlow.config_entry` really did land there. It just is not the newest thing this integration uses. `"homeassistant": "2025.2.0"` is the correct value.

## There is no way back in after a password change

`__init__.py` lines 138 to 139 wrap the whole update body in `except Exception as exc: raise UpdateFailed(...)`. A digest rejection comes out of `soap_client.py` line 37 as an `HTTPError` with code 401 and lands in that handler like any other failure. There is no `ConfigEntryAuthFailed` anywhere in the tree and no `async_step_reauth` in the config flow.

So when someone rotates the Innoxel account password, every entity goes unavailable and the coordinator keeps retrying the rejected credentials once a second, with no repair notification and no prompt to re-enter them. The only way out is to delete and re-add the entry.

Both `strings.json` and the four translation files already define `config.error.invalid_auth`, and no code path can ever set it: `config_flow.py` lines 66 and 109 collapse every exception to `cannot_connect`. Catching the 401 explicitly gives you the reauth trigger in the coordinator and makes that string reachable in the flow at the same time.

## Diagnostics freeze and keep reporting as if they were live

`__init__.py` lines 125 to 132 give the device-status fetch its own `try`/`except Exception` that logs a warning and continues, leaving `self._cached_devicestatus` holding the last good payload. Line 136 then republishes it, the outer update returns normally, and `last_update_success` stays true. Since nothing overrides `available`, all eight diagnostic sensors and the six supply binary sensors stay available and keep showing whatever voltages, temperatures and uptime were last read, for as long as that call keeps failing.

README line 101 sells this as "a failing diagnostics call never breaks the regular state updates", which is a fair goal, but the values freeze rather than going unknown, and nothing tells the user which they are looking at. Setting those entities unavailable when the cached payload is older than a couple of intervals would keep the good property and drop the misleading one.

## Thirteen hardcoded entity IDs, and some of them are constants

`sensor.py` 71, 92, 122, 157; `number.py` 71; `climate.py` 49; `binary_sensor.py` 87, 106, 129, 158, 181; `cover.py` 222; `switch.py` 88.

Assigning `self.entity_id` takes ownership of something the entity registry is meant to own, so a user's own rename does not survive. The sharper problem is that several of them carry nothing entry-specific:

```python
self.entity_id = "sensor.innoxel_diag_device_identity"
```

That is `sensor.py` line 122 in full. The weather and diagnostics IDs at 71, 92, 158 and 181 are the same shape, keyed only on a static suffix. A second Master 3 entry asks for IDs the first already holds for all 25 fixed entities. Your unique IDs are correctly scoped, so the registry will hand out `_2` variants rather than colliding, but the IDs you wrote quietly stop applying and anyone who automated against them gets a surprise. Drop the assignments and let `_attr_has_entity_name` and the registry generate them.

## Smaller items

- `InnoxelCover` declares `CoverEntityFeature.OPEN | CoverEntityFeature.CLOSE` at `cover.py` line 73 but implements `async_stop_cover` at line 162. Without `STOP` in `supported_features` Home Assistant will not dispatch that service to the entity, so the handler and its `_MOVE_TIMEOUT` fallback are unreachable. `InnoxelMotorG2Cover` declares it correctly at line 203. Also, the comment at line 17 says "90 s leaves margin" directly above `_MOVE_TIMEOUT = 65.0`.
- `SCAN_INTERVAL = 1` in `const.py` against a module-global `ThreadPoolExecutor(max_workers=2)` shared by every entry, with a 10 second per-request timeout. Every tenth tick also runs weather, timeswitch and per-module room climate, and `get_room_climate_state` sleeps 0.15 s per module in the loop. With nine modules that is over a second of deliberate sleeping plus nine sequential round trips inside a one second interval, which will trip Home Assistant's "update took longer than the scheduled interval" warning. Worth either raising the interval or moving the slow calls off the main tick.
- `InnoxelSoapClient.close()` is `pass`, so the `await client.close()` on unload does nothing, and the module-global executor is never shut down; its threads outlive every reload.
- `soap_client.py` line 31 logs the Innoxel username at debug level. It is half a credential pair and the action name alone would serve.
- Two README gaps: the "Diagnose Geräteinfo" sensor at `sensor.py` line 111 is shipped but absent from the entity table, and README line 30 still describes the plain output binary sensors as module index 45 and up, while `binary_sensor.py` lines 32 to 37 also create them for any In 8 / Out 8 module at any index, which is what the `v1.5.2` note says you intended.

Fix the floor and push a release and I will look again. The rest can ride along in the same round.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

prusuino:
Thanks for the review. All points addressed in [v1.6.0](https://github.com/prusuino/ha_innoxel_master3/releases/tag/v1.6.0); the PR body now links that release and its action runs.

- Floor raised to 2025.2.0 in `hacs.json` (`helpers.service_info.ssdp`).
- 401 now raises `ConfigEntryAuthFailed`; `async_step_reauth`/`reauth_confirm` added, `invalid_auth` is reachable from the setup and options flows, so the translation key is no longer dead.
- Polling split into two coordinators: outputs/dimmers/blinds stay on the 1 s tick, weather, time switches, room climate and device status moved to a 10 s coordinator. Measured on a Master 3: fast cycle 0.09-0.16 s, slow cycle 2.06-2.13 s, so each stays well inside its interval.
- Diagnostic and bus-supply sensors become unavailable after six failed slow polls (60 s) instead of showing the last value forever.
- All 13 `self.entity_id` assignments removed; the registry owns the ids and existing installations keep theirs. `has_entity_name` is deliberately not enabled: it would prefix every existing friendly name with the device name and effectively rename every entity in existing installations.
- Cover: `async_stop_cover` implemented to match the declared `STOP` feature.
- Client `close()` shuts down its executor; username removed from debug logging; the generic `except Exception` no longer swallows auth errors.

Re-requesting review.

PR #61, opened 2019-12-04
Adds new integration [DavidMStraub/homeassistant-homeconnect]

DavidMStraub:
Hi,

this is to add my integration of Home Connect:
https://github.com/DavidMStraub/homeassistant-homeconnect/
Thanks!

- [x] You are submitting only 1 repository.
- [x] You repository is compliant with https://hacs.xyz/docs/publish/start
- [x] You have tested it with HACS by adding it as a custom repository

PR #9431, opened 2026-07-22
Adds new integration [PraxiGEN/ha_qweather_pro]

PraxiGEN:
## Checklist

## Links

Link to current release: <https://github.com/PraxiGEN/ha_qweather_pro/releases/tag/v1.2.7>
Link to successful HACS action (without the `ignore` key): <https://github.com/PraxiGEN/ha_qweather_pro/actions/runs/33873091914>
Link to successful hassfest action (if integration): <https://github.com/PraxiGEN/ha_qweather_pro/actions/runs/33873091699>


frenck (changes requested):
Hi @PraxiGEN, thanks for the submission! 👋

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

frenck (changes requested):
Hi @PraxiGEN, thanks for merging master in. CI is green on both workflows, so here is the proper pass.

I reviewed against `v1.1.13`, the current stable release; `v1.2.0-beta.5` is flagged as a prerelease, so it is not what HACS would serve. The PR body still links `v1.1.10`, three stable releases back.

## Blocker: the coordinator swallows its own failure and serves stale data forever

`coordinator.py` lines 173 to 204. The `raise UpdateFailed("所有 API 抓取任务均失败")` at line 192 sits inside the `try:` that opened at line 173, and the `except Exception as err:` at line 194 catches it. That handler only adjusts `self.update_interval`, and then execution falls straight through to line 204, where the function parses `self._cache_data` and returns it.

So when every endpoint fails, `_async_update_data` returns normally. `last_update_success` stays `True`, `CoordinatorEntity.available` stays `True`, and the weather entity plus all five sensors keep serving the previous poll's readings. There is no bound on this: a broken key, an expired subscription or a dead network produces a weather card that looks perfectly healthy and is showing yesterday's temperature. The `update_time` attribute at line 295 even keeps advancing, so the one thing a user might check to notice actively lies to them.

Either move the `raise` outside the `try`, or narrow the handler to the exception types you actually mean to absorb.

## Nothing anywhere triggers reauth

There is no `ConfigEntryAuthFailed`, no `async_step_reauth` and no `reauth` string in any Python file. `api.py` lines 114 to 120 turn an HTTP 401 or 403 into an ordinary dict with `code: "401"`, and `coordinator.py` line 179 just treats that as a non-success endpoint.

Combined with the swallow above, a revoked or expired credential is completely indistinguishable from success: no repair issue, no failed entry, no unavailable entities, just permanent silent staleness. The config flow already classifies 401 and 403 correctly at lines 223 to 233, so the vocabulary exists; it is only the runtime path that is missing.

## Every user loads 871 KB of JavaScript on every frontend page

`__init__.py` lines 26 to 46 register the static path and then call `frontend.add_extra_js_url` for all three bundles unconditionally, gated only on the directory existing.

That is 830,971 bytes for `qweather-pro-card.js`, plus 30,072 and 9,917 for the other two, injected into every Home Assistant page load for everyone who installs the integration, whether or not they ever place the card on a dashboard.

And it collides. Your own issue #61 is a user on Home Assistant 2026.7.4 running exactly `v1.1.13`, the release proposed here, whose daily and hourly forecasts rendered blank. He isolated it himself by removing `apexcharts-card.js` from his Lovelace resources, after which everything worked, and concluded that QWeather Pro v1.1.13 conflicts with ApexCharts Card. That is correct: your bundle carries its own ApexCharts copy and is force-loaded globally, so it fights with one of the most widely installed cards in the ecosystem. The issue was closed with "解决问题就好" and no code change, and the collision is still in the release in front of me.

Gating the injection on the `custom_ui` option, or on the card actually being used, fixes both problems at once and is the thing I would most like to see before this is listed.

## The README describes the opposite of what ships

README line 23: "Frontend: Removes heavy libraries like Chart.js; uses native SVG to render temperature trend curves with 10× faster performance."

`qweather-pro-card.js` is 830,971 bytes and your own code starts at byte 817,397. The first 817 KB, 98 percent of the file, is a minified Lit plus ApexCharts vendor bundle; the string `apexcharts` occurs 1153 times in it. The release notes are honest about the direction, v1.1.11 says the forecasts now use curves, but the README still advertises having removed the library you added.

Line 13 makes the same claim about "ultra-fast native SVG weather cards".

## The documented card type does not exist

`md/DOCS.md` line 32 tells users to write:

```yaml
type: custom:qweather-card
```

The card registers as `qweather-pro-card`. Copy-pasting the documented YAML gives "Custom element doesn't exist: qweather-card".

The entity table in the same file has three more problems: it lists `sensor.qweather_warning_count` as "number of active weather alerts", but the shipped sensor's key is `warning_info` and `sensor.py` line 70 returns the alert title rather than a count, and neither the `warning_info` sensor nor the weather entity itself is documented. Five documented, six shipped, one of the five fictional.

## Licence notices

`LICENSE` line 3 reads `Copyright (c) 2026 hhh`. The repository history contains 100 commits authored by `dscao` and 6 by `shaonianzhentan`, and the initial commit is the latter's, so the notice does not reflect the authorship it covers. The README does credit `dscao/qweather` under Special Thanks, which is the right instinct; the licence file should match it.

Separately, the bundled Lit and ApexCharts ship with no licence text at all. I grepped the whole 831 KB file: zero occurrences of `@license` or `SPDX`, and no banner comment. Both libraries are MIT and both require their notices to be retained in redistributions.

## Smaller items

- `__init__.py` line 30 calls `os.path.exists` directly inside `async def async_setup_entry`. One stat per setup, so the impact is small, but it is exactly what Home Assistant's blocking-call detector flags.
- `api.py` line 99 calls `_generate_jwt()` on the event loop for every request, and that re-parses the PEM through `load_pem_private_key` each time. Ed25519 is cheap, but the key is being parsed dozens of times an hour instead of once. You already use `async_add_executor_job` correctly for keygen at `config_flow.py` line 154.
- `api.py` lines 68 to 75 give tenacity a `retry_error_callback` whose body is a `LOGGER.error(...)` call, so it evaluates to `None` and the function returns `None` instead of re-raising. `config_flow.py` line 206 then calls `.get()` on it. The broad `except` at line 243 turns that into `cannot_connect`, so it degrades rather than crashing, but it is wrong by construction.
- `coordinator.py` line 303, `air_raw.get("indexes", [{}])[0]`, raises `IndexError` on a present-but-empty list, and that call sits outside the try block. Line 348, `int(now_raw.get("windScale", 0))`, raises on a `None` or on a ranged value like `"1-3"`, and `windScale` is a string field upstream.
- `qweather-pro-more-info.js` line 165, `a.obs_time.slice(5, 16)`, has no null guard and throws during render in exactly the cold-start state the coordinator can produce.
- `__init__.py` lines 38 to 40 register the card before `qweather-pro-i18n.js`, but the card captures `window.QW_I18N` once at module evaluation. `add_extra_js_url` gives no ordering guarantee, so the card can permanently fall back to untranslated keys.
- `brand/logo备.png` is a 183 KB backup file with a non-ASCII name. Since `hacs.json` sets no `zip_release`, HACS downloads every file individually, and non-ASCII paths are a known source of trouble on some setups. It should not ship at all.
- `manifest.json` lists `["http"]` in dependencies but the integration also calls `frontend.add_extra_js_url`.

The rest holds up well: modern config and options flow patterns, correct reconfigure support, correctly scoped unique IDs, sensible dependency lower bounds that stay compatible across 2026.1 through 2026.8, twelve complete translations, and a clean sweep for dangerous constructs, unauthenticated views and secrets. Fix the coordinator and the global injection and I will look again.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

PraxiGEN:
Hi @frenck,

Thank you again for the thorough review. The release you examined (v1.1.13) predates a major rewrite of the integration; the current v1.2.7 addresses the issues you found:

**Blocker — coordinator swallowing failures:**&#8203; the broad `except Exception` handler now ends with an explicit `raise` after adjusting the interval, so `UpdateFailed` propagates as intended: `last_update_success` becomes `False`, entities go unavailable, and stale data is no longer served silently.

**Reauth:**&#8203; the runtime path now raises `ConfigEntryAuthFailed` on HTTP 401/403 responses and on local JWT generation failures, and `async_step_reauth` / `reauth_confirm` / `reauth_jwt` steps are fully implemented.

**Global JS injection:**&#8203; the frontend card was rewritten as a zero-dependency native SVG implementation (~32 KB — Lit is reused from Home Assistant itself, and ApexCharts is no longer bundled at all). Injection is now gated behind opt-in options ("Enable Custom UI Support" / "Replace native detail popup") instead of loading for every user, resources are unregistered on unload, and stale registrations are cleaned up by URL prefix.

**README:**&#8203; the claims now match reality — the card truly is native SVG with no bundled charting library.

**Documentation:**&#8203; the card type is corrected to `custom:qweather-pro-card`.

**Licence:**&#8203; `LICENSE` now carries the proper attribution line ("Copyright (c) 2026 PraxiGEN. Based on dscao/qweather by dscao, used under the MIT License."), and since no third-party bundles ship anymore, the retained-notice concern no longer applies.

**Smaller items:**&#8203; `os.path.exists` now runs via `async_add_executor_job`; the parsed PEM key object is cached instead of re-parsed on every request; the tenacity callback that returned `None` was removed; `indexes[0]` is guarded against empty lists; `windScale` goes through a safe numeric converter; the more-info script was fully rewritten (the unguarded `obs_time.slice` no longer exists); the `brand/logo备.png` backup file was removed; and `frontend` is declared via `after_dependencies`.

Thanks again — please take another look when you have time.

PR #9445, opened 2026-07-23
Adds new integration [shaxxx/dhmz-sea-temperature]

shaxxx:
## Checklist

## Links

Link to current release: <https://github.com/shaxxx/dhmz-sea-temperature/releases/tag/0.2.2>
Link to successful HACS action (without the `ignore` key): <https://github.com/shaxxx/dhmz-sea-temperature/actions/runs/30005325370/job/89199842112>
Link to successful hassfest action (if integration): <https://github.com/shaxxx/dhmz-sea-temperature/actions/runs/30005325370/job/89199842086>


frenck (changes requested):
Hi @shaxxx, thanks for the submission! 👋

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

shaxxx:
Thanks for the review! I've rebased `add-dhmz-sea-temperature` onto the latest `hacs/default` master — the conflict on the `integration` catalog is resolved and the branch is back to a clean one-line addition (head `a0982db8b60316abc80159456ed09a2789708761`). The validation workflows should re-run on the new head.

shaxxx:
@frenck All green, good to go.

frenck (changes requested):
Hi @shaxxx, thanks for the quick rebase back in August, and sorry for the wait on the real pass. Here it is.

I reviewed against `0.2.2`, which is the only release and matches what the PR body links. I also ran your parser against the live `vrijeme.hr/more_n.xml` feed to check the shape held up, and it does: 16 stations, six terms, and the `autom` attribute and empty `<Termin/>` cases you handle are both present in today's document. The card's escaping is careful and I found no XSS path in it, which is rarer than it should be.

Four things need to change.

## The integration writes into the user's Lovelace resource storage

`__init__.py` lines 44 to 66 reach into `hass.data["lovelace"].resources` and call `async_create_item` at line 64, or `async_update_item` at lines 59 to 61 when the URL has changed. Line 77 calls `async_delete_item` on entry removal.

Those mutate `.storage/lovelace_resources`, which is the user's dashboard configuration, and they do it through a collection that is not a public integration API. Setting up a config entry should not silently edit the resource list.

You already have the right thing sitting one line away: line 33 falls back to `frontend.add_extra_js_url`, which achieves the same load without touching storage. Please make that the only path. I understand the argument in your comment at lines 28 to 31 about the companion app, but the fix for that is to ship the card as a separate `plugin` submission, where the user knowingly adds the resource, rather than for the integration to own the list.

There is a second reason to drop it: line 38 is `getattr(hass.data.get("lovelace"), "resources", None)`, and line 39 guards on `hasattr`. If either internal is renamed, this degrades to the fallback in complete silence, so the user never learns why their card stopped loading.

## `hacs.json` declares no version floor at all

The whole file is:

```json
{
  "name": "DHMZ Sea Temperature",
  "render_readme": true
}
```

`config_flow.py` line 94 calls `self._get_reconfigure_entry()`. I checked core directly: that method is absent at 2024.7.0 and 2024.10.0 and first appears in 2024.11.0. Everything else you use is older, so `"homeassistant": "2024.11.0"` is the accurate floor. Without it HACS will offer this to any version, and a user on 2024.10 gets an `AttributeError` the moment they open Reconfigure.

## `available` is hardcoded to `True`, and the status sensor can lie forever

`sensor.py` lines 91 to 95, on `_BaseSensor`, so both entities inherit it:

```python
@property
def available(self) -> bool:
    return True
```

Your coordinator does the right thing, incidentally: the `raise UpdateFailed` at `coordinator.py` line 33 sits inside the `except` block rather than inside the `try`, so `last_update_success` genuinely goes false. Nothing then acts on it.

For `TemperatureSensor` the damage is bounded, because `native_value` returns `None` past the 24 hour retain window, so it degrades to unknown. That is a defensible design and your comment explains it.

`StatusSensor` has no such guard. Line 202 returns `s.status` straight off `self.coordinator.data`, and a `DataUpdateCoordinator` keeps the last successful payload indefinitely. If `vrijeme.hr` is down for a week, or the station disappears from the feed, `sensor.dhmz_sea_temperature_<x>_status` keeps reporting `ok` and keeps reporting itself available, with nothing to tell the user or their automations otherwise. That is the one sensor whose entire purpose is to say whether the reading can be trusted.

Either let `available` consult `last_update_success`, or give the status sensor the same age guard the temperature sensor has.

## The README documents a setup step that does not exist

README lines 30 and 31, install step 5: "Enter a device name, or leave it as the default (the station name), and click Finish."

The config flow schema has exactly two fields, `CONF_STATION` and `CONF_LANGUAGE`. There is no name field anywhere in `config_flow.py`, and the device name is set unconditionally to the station string at `sensor.py` line 66. Either add the field or drop the step.

## Smaller items

- `sensor.py` lines 114 to 116 and 196 to 198 assign `self.entity_id` in the constructor while `_attr_has_entity_name = True` is set, which is the mechanism meant to derive it. `async_generate_entity_id` only consults `hass.states`, and at construction time on a restart the entity is not in the state machine yet, so its collision avoidance is not doing anything for you. The registry wins for already registered entities, so this is mostly inert rather than harmful. Line 113 also sets `_attr_name` from a hardcoded dictionary rather than a `translation_key`, which the status sensor does correctly at line 187.
- The per instance `language` option duplicates Home Assistant's own UI language and translation machinery. It works, but it means the same install can show two languages depending on which entity you look at.
- `sensor.py` lines 152 and 153 put the full `readings` dict and `hours` list into `extra_state_attributes`, so the recorder persists them on every state change. Six terms is small, but it is also what forces the card's `_sig` change detection at line 228.
- Seven of the eight shipped attributes are undocumented; the README lists only `automatic_buoy`.
- One commit, one release, and nothing since 2026-07-23. That is not disqualifying, but it does mean there is no track record to read, so the code has to carry the whole case.

Cross linking `kpisacic/DHMZ-home-assistant-custom-component` in both the README and the release notes, and being explicit about the scope difference, was the right thing to do and I noticed it.

Drop the resource writes, add the floor, fix the status sensor and the README step, and push a release. I will look again.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

shaxxx:
Thanks for the thorough review, @frenck — all four are addressed in 0.3.0:

Resource writes removed. The integration no longer touches hass.data["lovelace"].resources or any dashboard storage — __init__.py is sensors-only. The card is now a separate HACS plugin, shaxxx/dhmz-sea-temperature-card.
Version floor: "homeassistant": "2024.11.0" in hacs.json (for _get_reconfigure_entry()).
Status sensor: dropped the hardcoded available = True; StatusSensor now inherits last_update_success, so it goes unavailable when the feed is down instead of reporting ok indefinitely (with regression tests for the feed-down and station-missing cases). TemperatureSensor keeps its 24 h-retain design.
README: removed the non-existent "device name" step, documented all sensor attributes, and moved the card instructions to the plugin.
I left the smaller items (constructor entity_id, per-instance language, readings/hours in attributes) as-is — the entity_id and language behaviours are intentional, and the card needs readings/hours. Happy to revisit if you'd prefer.

Integration release: https://github.com/shaxxx/dhmz-sea-temperature/releases/tag/0.3.0

The card now has its own plugin submission: hacs/default#10510. Since it's the exact card code you already looked at here (you noted the escaping was careful, no XSS path), would you be able to take #10510 alongside this one? It'd spare it a separate wait in the queue — no problem if it needs the normal flow.

One consistency question, purely so the requirement is applied evenly: timmaurice/sea-temperatures (merged, in default) registers its card through the same hass.data["lovelace"].resources API — is integration-side auto-registration meant to be disallowed across the board? Either way I've moved to the plugin approach here; just flagging it.

PR #9457, opened 2026-07-23
Adds new integration [tieskuh/veloretti-ble-ha]

tieskuh:
## Checklist

## Links

Link to current release: <https://github.com/tieskuh/veloretti-ble-ha/releases/tag/v0.3.2>
Link to successful HACS action (without the `ignore` key): <https://github.com/tieskuh/veloretti-ble-ha/actions/runs/29991158774/job/89154098195>
Link to successful hassfest action (if integration): <https://github.com/tieskuh/veloretti-ble-ha/actions/runs/29991158774/job/89154098167>

### DIFF
--- a/integration
+++ b/integration
@@ -2822,6 +2822,7 @@
   "tiagonmas/recbms",
   "tibuntu/homeassistant-kubernetes",
   "tiehfood/ha-festool-connected",
+  "tieskuh/veloretti-ble-ha",
   "Tiimber/ha-smhi-odp",
   "tijsverkoyen/HomeAssistant-FusionSolar",
   "tikismoke/home-assistant-plcbus",
--- a/integration
+++ b/integration
@@ -2946,6 +2946,7 @@
   "tiagonmas/recbms",
   "tibuntu/homeassistant-kubernetes",
   "tiehfood/ha-festool-connected",
+  "tieskuh/veloretti-ble-ha",
   "Tiimber/ha-smhi-odp",
   "tijsverkoyen/HomeAssistant-FusionSolar",
   "tikismoke/home-assistant-plcbus",


frenck (changes requested):
Hi @tieskuh, thanks for the submission! 👋

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

frenck (changes requested):
Hi @tieskuh, thanks for the rebase; the conflict is cleared and it merges clean now. Here is the proper pass.

I reviewed against `v0.3.2`. Seven entities documented in the README table and seven shipped, with the enabled-by-default column matching exactly. Release notes all check out against code. The `hacs.json` floor of `2024.8.0` is correct and slightly conservative, since `entry.runtime_data` puts the real floor at 2024.6.0. The `proto/` directory documenting how you derived the protocol is a genuinely nice thing to ship. Nothing in the security sweep came back positive.

Two things, and both come down to the same shape: a bike that is failing looks like a bike that is asleep.

## A persistently failing bike is indistinguishable from a healthy one

`coordinator.py` lines 132 to 156. The poll body catches `ComoduleAuthError` at line 146 and logs a warning, catches `BleakError` at line 149 and logs at debug, and then line 156 is `return self.data` on every path including both failure paths. Nothing propagates.

Core's `ActiveBluetoothDataUpdateCoordinator` sets `last_poll_successful = False` and logs the failure only when the poll method raises, so with nothing raised that flag stays `True` forever and the built in failure and recovery reporting never fires.

That would be survivable on its own, except for what sits on top of it. `entity.py` lines 47 and 48 short-circuit `available` to `True` for anything carrying `always_available`, which is battery, last seen and connectivity. And `binary_sensor.py` line 47 computes Connectivity as `coordinator.available or coordinator.streaming`, where `coordinator.available` is core's passive coordinator property meaning only "an advertisement was seen recently".

So a bike whose auth handshake is rejected every time still advertises, so Connectivity reads on, and battery keeps showing whatever it read before, and neither ever goes unavailable. The README at lines 36 and 37 sells Connectivity as the entity that "tells you whether the reading is live or the last-known one", and in this case it says live while the reading is not.

Credit where it is due: `coordinator.py` lines 163 and 164 only advance `last_successful_poll` when a register actually answered, so Last seen does correctly stop moving. That is the one honest signal, and it is why I rate this worth fixing rather than serious. Letting `ComoduleAuthError` propagate would be enough on its own, since that is the failure mode a user cannot recover from without knowing about it.

## The README promises debug output that no code produces

README lines 178 and 179: "The debug log includes the raw register packets, which is exactly what's needed to map a new model", with the logger YAML at lines 171 to 176.

There are eight `_LOGGER` calls in the component, at `comodule.py` line 199 and `coordinator.py` lines 139, 147, 149, 155, 208 and 222, and every one logs an exception string. None logs a packet. The packets do exist, at `comodule.py` line 124 where `data.raw[key] = packet.hex(" ")`, but that field is never logged, never exposed as an attribute, and there is no diagnostics platform to surface it.

This matters more than a typical doc slip because it is also the documented route for anyone wanting to add a new bike model, and following it produces nothing useful. Either log `data.raw` at debug or add a `diagnostics.py` that returns it.

## Smaller items

- `proto/probe.py` lines 17 to 19 look like real BLE MAC addresses in the usage examples, most likely your own bike. That directory never reaches a user, since `content_in_root: false` means HACS installs only the component, so this is privacy housekeeping rather than a defect. I mention it because `.gitignore` lines 7 and 8 show you were already thinking about not publishing them.
- README line 4 still lists "charging" among the readings, while lines 80 to 84 explain plainly that there is no charging sensor and why. Stale copy in the opening sentence.
- `manifest.json` line 18 requires `bleak-retry-connector>=3.6.0` with no upper bound. Core already ships and pins that package, so in practice nothing is installed and the risk is theoretical, but for a bluetooth integration declaring it at all is arguably unnecessary.
- `manifest.json` line 16 says `iot_class: local_polling`, while from v0.3.0 the design is really push over notify while connected, with polling as the outer loop. Not wrong, just no longer the whole picture.
- `custom_components/veloretti_ble/brand/icon.png` is unused; Home Assistant reads brand assets from the `home-assistant/brands` repository, not from inside the component.

Let the auth error propagate, and either produce the debug output the README promises or change what it says. Push a release after that and I will look again.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

tieskuh:
Thanks for the review. Both blocking findings were correct and are fixed in v0.4.1.

**Auth failures now surface.** `ComoduleAuthError` propagates, so the base coordinator logs it and clears `last_poll_successful` until a later poll recovers. `BleakError` stays at debug deliberately: that one is the bike falling asleep or drifting out of range mid-poll, which would otherwise error constantly on a bike that is off most of the time. Connectivity is now `(available or streaming) and not auth_failed`, so it no longer reports a stale reading as live.

**Debug output now exists.** Raw packets are kept for every register the module answers, including ones this decoder cannot decode yet, keyed by register id, and logged on each poll and each notifier push. A new diagnostics platform also offers them as a download.

**README opening line.** Fixed, it no longer mentions charging.

**`iot_class`.** Now `local_push`.

**`bleak-retry-connector`.** Left declared, since the integration imports it directly and core's pin wins in practice either way. Happy to drop it if you prefer.

**`proto/probe.py` addresses.** These are placeholders: all three examples read `E1:23:45:67:89:AB`. A `git grep` over the repository finds no real address in it.

**`brand/icon.png`.** Still in use. Since 2026.3, Home Assistant serves brand images via `/api/brands/...`, and `homeassistant/components/brands/__init__.py` builds `brand_dir = Path(integration.file_path) / "brand"` and tries those local files before the CDN. The `home-assistant/brands` README now directs custom components to ship icons directly, and the HACS action's brands check passes against the local file.

Release: https://github.com/tieskuh/veloretti-ble-ha/releases/tag/v0.4.1

PR #9475, opened 2026-07-24
Adds new integration [LuckyTriple7/ha-beach-weather]

LuckyTriple7:
## Checklist

## Links

Link to current release: https://github.com/LuckyTriple7/ha-beach-weather/releases/tag/v1.1.0
Link to successful HACS action (without the `ignore` key): https://github.com/LuckyTriple7/ha-beach-weather/actions/runs/33271763667
Link to successful hassfest action (if integration): https://github.com/LuckyTriple7/ha-beach-weather/actions/runs/33271763644

## Related

The bundled Lovelace card has been split out into its own repository and is submitted separately as a plugin: hacs/default#10451

### DIFF
--- a/integration
+++ b/integration
@@ -1724,6 +1724,7 @@
   "lrehmann/pihole_presence",
   "lsellens/thesimple-thermostat",
   "LucasTor/FoxESS-T-series",
+  "LuckyTriple7/ha-beach-weather",
   "LuckyTriple7/ha-strato-dyndns",
   "ludodefgh/montreal-snow-removal",
   "ludvikroed/homely-integration",
--- a/integration
+++ b/integration
@@ -1788,6 +1788,7 @@
   "lrehmann/pihole_presence",
   "lsellens/thesimple-thermostat",
   "LucasTor/FoxESS-T-series",
+  "LuckyTriple7/ha-beach-weather",
   "LuckyTriple7/ha-strato-dyndns",
   "ludodefgh/montreal-snow-removal",
   "ludvikroed/homely-integration",


frenck (changes requested):
Hi @LuckyTriple7, thanks for the submission! 👋

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

frenck (changes requested):
Hi @LuckyTriple7, thanks for the rebase. The conflict is cleared and both HACS and hassfest are green on the current release, so here is the proper pass.

Two notes before the findings. The PR body still links `v0.5.0` and the action runs from back then, while the repository is on `v0.24.0`; please update it. And I reviewed against `v0.24.0`, where the entity accounting is exact: 37 documented in the README table and 37 shipped, no undocumented entity and nothing documented that does not exist.

## Blocker: the declared floor is about twenty-five releases below what the code needs

`hacs.json` declares `"homeassistant": "2024.1.0"`. `__init__.py` line 65 reads `lovelace.resource_mode`.

I counted occurrences of `resource_mode` in core's `homeassistant/components/lovelace/__init__.py`: zero at 2026.1.0, six at 2026.2.0. The field was `mode` before that and only became `resource_mode` in 2026.2.0. The compatibility shims on `LovelaceData` cover subscript and `.get()` access, not attribute access, so on 2026.1.0 this raises `AttributeError` inside the `async_at_started` callback rather than degrading.

There are three more constraints below it, all above your declared floor: `LOVELACE_DATA` in `lovelace/const.py` is absent through 2025.1.0 and first present at 2025.2.0, and the import at line 62 is not guarded, so below that it is a hard `ImportError`; `StaticPathConfig` needs 2024.7.0; and the implicit `OptionsFlow.config_entry` you read at `config_flow.py` lines 207, 242 and 265 needs 2024.12.0.

So the true floor is `2026.2.0`. Everything between the declared 2024.1.0 and 2025.1.x fails at import, and everything from 2025.2.0 to 2026.1.x fails with `AttributeError`.

## Blocker: the integration writes the user's Lovelace resource store and never cleans up

`__init__.py` lines 51 to 85 reach into `hass.data[LOVELACE_DATA]`, force the store to load with `resources.async_get_info()`, iterate `resources.async_items()`, and then call `resources.async_update_item(...)` or `resources.async_create_item(...)`.

That is the shared, user-owned `.storage/lovelace_resources`. A config entry setting itself up should not be editing it.

Your comment at lines 52 to 60 makes a real argument, and I want to acknowledge it: `frontend.add_extra_js_url` does get rendered behind an `isModern` user-agent check, and on a browser that check rejects, the card silently fails to load. That is a genuine problem and you diagnosed it correctly. But the answer is to ship the card as a separate `plugin` submission, where HACS writes the resource as the package manager acting on the user's behalf and the user knows it happened, rather than to have an integration write it behind their back.

There is a second, independent problem with the current approach. The URL carries `?v={CARD_VERSION}`, so lines 79 to 82 rewrite the entry on every version bump, and there is no removal path at all: no `async_remove_entry` anywhere in the tree and no `async_delete_item` call. Uninstalling the integration leaves an orphaned resource pointing at a dead `/beach_weather_static/...` URL, which Lovelace then tries to load on every dashboard render, forever, on an installation that no longer has the integration.

## All 37 entities hardcode their entity ID

`sensor.py` lines 192, 894, 982, 1002 and 1062, `button.py` line 41, `weather.py` line 171.

The comment at `sensor.py` lines 971 and 972 says this is to keep existing picture-elements cards working, so I know it is deliberate. It still takes ownership of something the entity registry is meant to derive, it means a user's own rename does not survive, and it sits oddly next to `_attr_has_entity_name = True` on the same classes, which is the mechanism that would otherwise do the naming.

## Smaller items

- The card editor interpolates registry names into `innerHTML` without escaping: `beach-weather-card.js` lines 703 to 710 for the device name and lines 825 to 833 for the entity friendly name. Both trace back to the user's own typed location name, so this is self-inflicted rather than remotely reachable, and every value that renders live entity state correctly uses `textContent`. Worth closing anyway, since it is one helper away. Same shape at lines 733 to 737, where the user's custom background URL goes straight into a CSS `url("...")` with no quote escaping.
- README line 165 says other HTTP errors back off for five minutes. `const.py` line 57 is `ERROR_BACKOFF = {403: 1800, 429: 900, 503: 30}`, so 503 is thirty seconds. The 503 case was added in v0.22.1 and made it into the CHANGELOG but not the README.
- `__init__.py` line 30 keeps `CARD_VERSION` in sync with `manifest.json` by hand, with a comment as the only enforcement. They agree today; if they ever drift, browsers serve a cached card.
- `brand/icon.png` and `icon@2x.png` sit in a `brand/` subdirectory that neither Home Assistant nor HACS reads below 2026.3, and the bundled photos are about 1.8 MB of the download. Worth a thought on payload size.
- Diagnostics redacts latitude and longitude, which is the right instinct, but the beach name and slug in `entry.data` are not redacted and `coordinator.data` goes out raw.

One thing worth flagging that is not a defect: core Home Assistant ships an `open_meteo` weather integration against the same Forecast API, so the weather entity and the atmospheric sensors overlap it. The marine half, the twelve wave and swell sensors plus the bathing and surf scoring, is genuinely additive and core has nothing like it, and your README is upfront about the split. I mention it only so you know the overlap is understood and is not what is holding this up.

Fix the floor and the resource writes and push a release, and I will look again.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

LuckyTriple7:
Thanks — all of it is addressed. The PR body now points at v1.1.0.

**The card is out of the integration.** It ships as its own dashboard repository,
LuckyTriple7/ha-beach-weather-card, submitted separately in #10451. No reaching into
hass.data[LOVELACE_DATA], no resource writes: HACS registers it as the package manager,
and uninstalling it removes the resource again.

**The floor.** Dropping the card dropped the two constraints it carried — LOVELACE_DATA
(2025.2) and resource_mode (2026.2) — and the static path took StaticPathConfig with it.
What is left is the implicit OptionsFlow.config_entry, so hacs.json now declares
2024.12.0. Thanks for pinning the resource_mode rename to the release; I had missed that
the field was renamed at all.

**The orphaned resource.** A one-time cleanup at start deletes resources under the old
/beach_weather_static/ prefix and nothing else, then the integration never touches the
store again. It reads through hass.data["lovelace"] rather than importing LOVELACE_DATA,
so the cleanup does not reintroduce the 2025.2 floor. Seven tests cover it, including
YAML resource mode and a failing cleanup not taking setup down with it.

**The hardcoded entity IDs are gone.** All seven assignments removed; the registry
derives them from the device and entity name now, so a new location gets
sensor.platja_de_muro_water_temperature. Existing entities keep the IDs the registry
already holds, so no dashboard or automation breaks on update.

Two notes on that one, since I looked into it properly. Setting entity_id in the
constructor turned out to be a suggestion rather than ownership — entity_platform reads
it as suggested_object_id and then assigns entity.entity_id = entry.entity_id, so a
user's rename did survive; I have a test that renames a sensor, reloads and asserts it.
The reason to change it anyway is the second thing I found: the card looked its entities
up by the sensor.<key>_<slug> ID prefix, which *did* break on a rename. It now resolves
them through the registry translation_key, so it handles the old IDs, the new IDs and
renamed entities alike. That fix is card 0.2.0 and it is what actually made the ID
change safe.

**Smaller items.** The editor escapes registry names into innerHTML and the custom
background URL into the CSS url(). The README says 30s for 503. Diagnostics redact the
location name, its slug and the raw coordinator payload. CARD_VERSION went with the card;
it is versioned by its own release now. The photos are WebP at 1024px and the default
background is no longer a byte-identical second copy of the partly-cloudy one — 1.79 MB
down to 486 KB in the card repo, and this repo's download is about 1.8 MB smaller.

**One thing I kept: brand/.** hacs/action's brands validator looks for exactly
<content path>/brand/icon.png and only falls back to the home-assistant/brands repository
when it is missing, so removing the directory failed the check. It is 40 KB — the 1.8 MB
was the photos.

And thanks for the note on the open_meteo overlap — understood, and the README stays
explicit about which half is additive.

PR #9491, opened 2026-07-25
Adds new integration [AboveColin/HA-Liberty-Global-Gateway]

AboveColin:
## Checklist

## Links

Link to current release: <https://github.com/AboveColin/HA-Liberty-Global-Gateway/releases/tag/v2.0.3>
Link to successful HACS action (without the `ignore` key): <https://github.com/AboveColin/HA-Liberty-Global-Gateway/actions/runs/33272568476>
Link to successful hassfest action (if integration): <https://github.com/AboveColin/HA-Liberty-Global-Gateway/actions/runs/33272568431>

### DIFF
--- a/integration
+++ b/integration
@@ -30,6 +30,7 @@
   "abnvle/ha-imgw-pib-monitor",
   "abnvle/ha-matomo",
   "AboveColin/HA-Forgejo",
+  "AboveColin/HA-Liberty-Global-Gateway",
   "AboveColin/HA-Philips-Pet-Series",
   "ABovsh/eveus",
   "absent42/Aqara-Advanced-Lighting",
--- a/integration
+++ b/integration
@@ -32,6 +32,7 @@
   "abnvle/ha-imgw-pib-monitor",
   "abnvle/ha-matomo",
   "AboveColin/HA-Forgejo",
+  "AboveColin/HA-Liberty-Global-Gateway",
   "AboveColin/HA-Philips-Pet-Series",
   "AboveColin/HA-Traefik",
   "ABovsh/eveus",


frenck (changes requested):
Hi @AboveColin, thanks for the submission! 👋

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

frenck (changes requested):
Hi @AboveColin, thanks for the rebase. Here is the proper pass.

Reviewed against `v2.0.0`, which is what the PR body links and the only release. The `liberty-global-gateway` package is real, published and exact-pinned, and every translation key the code uses exists with none orphaned. Diagnostics has a thorough `TO_REDACT` and even regex-scrubs MAC addresses out of the event log, which is more care than most submissions take. The socket call in the config flow is correctly executor-wrapped, and there is a working reauth flow.

Four things.

## `hacs.json` declares no version floor

The file is three keys and none of them is `homeassistant`. `config_flow.py` lines 17 to 21 import `SsdpServiceInfo`, `ATTR_UPNP_SERIAL` and `ATTR_UPNP_UDN` from `homeassistant.helpers.service_info.ssdp`. That module 404s at core 2025.1.4 and first appears at 2025.2.0; I checked both tags. So the accurate floor is `"homeassistant": "2025.2.0"`, and without it HACS offers this to versions where the config flow cannot import.

While you are in that file, the `name` is 148 characters listing five model numbers. That string is what a user sees in the HACS list, and it will wrap or truncate badly. "Liberty Global Cable Gateway" would serve better, with the model list in the README where it already lives.

## A busy session leaves every entity showing stale data indefinitely

`__init__.py` lines 124 to 129:

```python
except GatewaySessionBusyError as err:
    if self.data is not None:
        _LOGGER.debug("Gateway session busy; keeping previous data: %s", err)
        return self.data
```

Returning normally from `_async_update_data` means `last_update_success` stays true, so `available` at `entity.py` line 360 stays true and `system_health` keeps reporting ok. If someone leaves the router web UI open, every one of the 55 entities serves values from an arbitrarily old poll for as long as that lasts, with nothing but a debug line to say so.

I understand the intent, and your README explains it: flapping 55 entities to unavailable on a transient contention is worse than holding. But this is unbounded. Bounding it, replaying for a couple of consecutive cycles and then raising `UpdateFailed`, keeps the good property and drops the misleading one.

The same shape sits at lines 79 to 88 for the hosts read, where the previous host list is reused while the rest of the payload is fresh, so device tracker presence silently lags the other entities. That one does correctly re-raise when there is no previous data.

## A lockout is treated as an authentication failure

`__init__.py` lines 122 and 123 group `GatewayLockoutError` with `GatewayAuthError` into `ConfigEntryAuthFailed`. That tears down the entry and raises a "reauthentication needed" repair asking for credentials.

But your own README says a lockout is a few-minute cooldown that clears by itself with the correct password. So the user is told their credentials are wrong when they are not, and re-entering the same correct password is the documented remedy for a problem that was going to resolve on its own. Your config flow already distinguishes the two, with a separate `lockout` error string. `UpdateFailed`, or a backoff, is the right response at the coordinator.

## Five unserialised sessions against a device you document as single-session

There is no `asyncio.Lock` or `Semaphore` anywhere in the tree. Login and logout pairs appear at `__init__.py` lines 67 and 137, `button.py` 52 and 57, `number.py` 62 and 67, `switch.py` 94 and 101, and `config_flow.py` 108 and 124.

Your README says the gateway allows exactly one authenticated session. So pressing reboot, moving the LED brightness slider or flipping a switch while a poll is in flight opens a second login on the same client, and its `finally: logout()` revokes the token the coordinator is holding. Entity commands and polls can break each other, and the failure surfaces as exactly the session-busy error the code above then swallows. A single lock around the login-to-logout span would remove both halves of this.

## Smaller items

- The README covers 42 of the 55 shipped entities. Missing entirely: `dhcp_pool`, `mac_filters`, `port_triggers`, `ip_port_filters`, `wifi_2g_channel`, `wifi_5g_channel`, `wps_2g`, `wps_5g`, `dslite`. And four more are described as attributes of the WAN IP sensor when they ship as standalone entities: `dns_servers`, `wan_gateway`, `wan_ipv6`, `wan_lease_time`. Nothing is documented that does not exist, so the gap is one-directional.
- `device_tracker.py` line 495 sets `_attr_unique_id`, but `ScannerEntity.unique_id` is a property override returning `self.mac_address`, so that line has no effect and the real unique ID is the bare MAC. Harmless today because both flow steps abort on a second entry, but the line is misleading.
- `diagnostics.py` lines 41 to 45: `_model_dump` redacts only when the object exposes a `.raw` dict and otherwise falls through to `str(obj)`, which for a dataclass is its full repr with no redaction. That path covers `lan`, `ipv6`, `registration`, `system`, `cable`, `modem_mode`, `software_update` and each wifi config. I could not confirm whether every model in the library carries `.raw`, so I am not calling it a leak, but the fallback is unsafe by construction.
- `sensor.py` lines 28 to 34 recompute the boot timestamp from uptime on every poll, so the uptime timestamp sensor jitters by up to a second every five minutes and churns recorder history. Deriving it once and caching would settle it.
- `brand/icon.png` and `brand/logo.png` are byte identical, as are the `@2x` pair, and `home-assistant/brands` has no entry for this integration, so none of the 14.7 KB is doing anything.

Add the floor, bound the stale replay, reclassify the lockout, and serialise the sessions. Push a release and I will look again.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

AboveColin:
Fixed in [v2.0.3](https://github.com/AboveColin/HA-Liberty-Global-Gateway/releases/tag/v2.0.3).

- `hacs.json` declares `"homeassistant": "2025.2.0"`.
- The session-busy replay is bounded at 2 cycles (10 minutes), then `UpdateFailed`. The hosts fallback got the same bound.
- A lockout raises `UpdateFailed` instead of `ConfigEntryAuthFailed`.
- The coordinator owns an `asyncio.Lock` behind a `session()` context manager covering login to logout. The coordinator, button, number and switch all take it.
- Smaller items: the boot timestamp is derived once, `_model_dump` walks dataclass fields through `async_redact_data`, the dead `_attr_unique_id` is gone, and the README covers all 55 entities.

On the brand: `logo.png` and `logo@2x.png` are gone, since core falls `logo.png` back to `icon.png`. `brand/icon.png` stays. Since 2026.3 a custom integration bundles its images there, `loader.py` tests for exactly that folder in `Integration.has_branding`, and `home-assistant/brands` auto-closes custom-integration PRs. The missing brands entry is expected.

PR #9492, opened 2026-07-25
Adds new integration [somansch/lufop_radar]

somansch:
## Checklist

## Links

Link to current release: https://github.com/somansch/lufop_radar/releases/tag/v2.2.0
Link to successful HACS action (without the `ignore` key): https://github.com/somansch/lufop_radar/actions/runs/33277035490/job/99165448615
Link to successful hassfest action (if integration): https://github.com/somansch/lufop_radar/actions/runs/33277035490/job/99165448707


frenck (changes requested):
Hi @somansch, thanks for the submission! 👋

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

somansch:
> Hi @somansch, thanks for the submission! 👋
> 
> This PR currently has a merge conflict with `master` on the `integration` catalog file. Other entries have been added near the insertion point since this PR was opened, so the simple one-line addition can no longer be auto-merged. The HACS validation actions also haven't rerun on the current head because of it.
> 
> Please rebase the branch on the latest `hacs/default` master (or merge master in) so the conflict resolves and the validation workflows fire again. Once CI is green I'll do the proper review pass.
> 
> ../Frenck
> 
> [](https://www.linkedin.com/in/frenck/)    [](https://youtube.com/@frenck)    [](https://bsky.app/profile/frenck.social)    [](https://fosstodon.org/@frenck)    [](https://www.instagram.com/frenck/)    [](https://www.threads.net/@frenck)    [](https://www.facebook.com/frenck.dev/)    [](https://x.com/frenck)    [](https://www.tiktok.com/@frenck.nl)
> 
> Blogging my personal ramblings at [frenck.dev](https://frenck.dev)

Hi @frenck , done, rebased on master

somansch:
> **Your branch seems out of date.** Please update your branch with the latest changes from the base branch.

done

frenck (changes requested):
Hi @somansch, thanks for the rebase back at the start of August. Here is the proper pass.

I reviewed against `v2.1.2`, which is the current release; the PR body still links `v2.0.0`, and the action run it points at predates two releases. The documentation is unusually precise: I counted every documented entity, attribute, event field, service field and config option against what ships and every one matched exactly, twelve attributes, ten event fields, eleven config fields, and the route-corridor sample-point arithmetic in the README even works out to the 21 points the code produces.

Three things.

## `hacs.json` declares no version floor

The whole file is four keys and none of them is `homeassistant`:

```json
{
  "name": "Lufop Radar",
  "render_readme": true,
  "zip_release": true,
  "filename": "lufop_radar.zip"
}
```

`config_flow.py` line 15 imports `section` from `homeassistant.data_entry_flow` at module level and uses it at lines 60 and 75. I counted `class section` in core's `data_entry_flow.py`: zero at 2024.6.0, one at 2024.7.0. So the accurate floor is `"homeassistant": "2024.7.0"`, and without it HACS offers this to versions where the config flow raises `ImportError` at module load with no actionable message.

The `zip_release` and `filename` pairing is right, incidentally: the release asset is named exactly `lufop_radar.zip` and the workflow builds it with the manifest at the archive root, which is what that flag needs.

## The radar markers keep showing stale positions through an outage

`LufopLocationEvent` at `geo_location.py` line 134 extends `GeolocationEvent` directly and holds the coordinator by hand at line 140. It is not a `CoordinatorEntity`, it defines no `available`, and nothing in the file consults `last_update_success`.

When a poll fails, the coordinator keeps its previous `self.data` and still notifies listeners, so `_sync_entities` runs against stale data and every radar marker goes on reporting its last known position with `available` still true. Meanwhile `LufopTotalSensor` is a `CoordinatorEntity` and does the right thing, so during an outage the count sensor goes unavailable while the markers stay on the map looking live. For a speed-camera integration people use while driving, that is the wrong way round.

The coordinator itself is correct and I checked it specifically: the `raise UpdateFailed` at `coordinator.py` line 116 is inside the only handler, nothing swallows it, and no failure path returns cached data. It is purely the entity's availability wiring.

## A revoked API key never asks for a new one

There is no `ConfigEntryAuthFailed` anywhere and no `async_step_reauth` in the config flow, whose steps are exactly `user`, `area`, `waypoint`, `route_corridor` and `route_options`. A rejected key comes back as HTTP 4xx, becomes `LufopAPIError`, and `coordinator.py` lines 115 and 116 turn that into `UpdateFailed`. So the coordinator retries forever and the user is never prompted. On first setup it surfaces as `ConfigEntryNotReady` and retries forever there too.

You do surface Lufop's own message text rather than a bare status code, which is better than most, so the vocabulary to distinguish the cases is already there.

## Smaller items

- `sensor.py` line 25 is `class LufopTotalSensor(CoordinatorEntity)`, added to the sensor platform but not a `SensorEntity`. It overrides `state` rather than implementing `native_value`, and line 52 puts `{"state_class": SensorStateClass.MEASUREMENT}` into `extra_state_attributes` instead of setting `_attr_state_class`. Set that way the recorder's statistics machinery ignores it, so the entity asks for long-term statistics and does not get them.
- Unique IDs key on the user-chosen display name (`geo_location.py` line 145, `sensor.py` line 37) rather than `entry.entry_id`. The duplicate-name abort you added in v2.1.2 does close the collision, and I could not construct a live one, but it is a guard bolted onto the wrong ID scheme, and the entity-registry purge at `geo_location.py` lines 97 to 106 now depends on that prefix too.
- `LufopOptionsFlow.__init__` at `config_flow.py` line 344 overrides the parent and never calls `super().__init__()`, so `self._options` is never set and `self.options` would raise. Nothing reads it today, so it is inert, but it is a trap for the next edit. Related: `OptionsFlowWithConfigEntry` still exists at 2026.8.0 but is documented as kept only for backward compatibility with custom integrations, so it is worth moving off.
- `fr.json` is missing `config.abort.already_configured`, which `en.json` and `de.json` both have. That is the abort you added in v2.1.2, so a French-locale user typing a duplicate name sees the raw key.
- `geo_location.py` lines 36 to 52 interpolate `radar["vitesse"]`, straight from the API, into generated SVG markup and then percent-encode the whole document. Encoding the document uniformly does not escape the value relative to the markup around it, so a value containing `</text>` breaks out of the element. The sink is `entity_picture`, where scripts do not execute, so this is low, but `xml.sax.saxutils.escape` before interpolation would close it.
- `brand/icon.png` inside the component is 25 KB that Home Assistant never reads below 2026.3; brand assets live in `home-assistant/brands`.

On the overlap with your own `blitzer` integration, which is already listed: different upstream, different domain, and the route-corridor search has no counterpart there, so I am treating these as two integrations rather than a resubmission. Worth saying explicitly so it does not come up again.

Add the floor, make the markers honour `last_update_success`, and wire up reauth. Push a release after that and I will look again.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

somansch:
Thanks — this is the most careful review this repo has had, and three of the findings were exactly right. All of it is in [v2.2.0](https://github.com/somansch/lufop_radar/releases/tag/v2.2.0), and the PR body above now points at that release and at a run from it.

## The version floor

Added, `"homeassistant": "2024.7.0"`. Your reading of `section` matches mine.

## Availability

`LufopLocationEvent` is a `CoordinatorEntity` now, and takes its own record out of each poll instead of being handed one. `_sync_entities` returns early when the poll failed, so nothing is added, removed or announced off data that did not arrive.

Measured on a live instance, with the failure injected at the API boundary:

| | markers | count sensor | entry |
|---|---|---|---|
| healthy | 12, each with a distance | 12 | loaded |
| poll fails | 12 × unavailable | unavailable | loaded |
| next success | 12, each with a distance | 12 | loaded |

## Reauth — the conclusion was right, the premise was not

Before wiring it up I checked what a rejected key actually does. The radar endpoint does not look at the key at all:

```
?key=definitely-not-a-valid-key&format=json&pays=FR  → 200, real radars
?key=&format=json&pays=FR                            → 200, real radars
?format=json&pays=FR      (no key parameter at all)  → 200, real radars
?key=definitely-not-a-valid-key&format=json&pays=DE  → 200, real radars
```

So a revoked key would never have produced a 4xx. It would have gone on working, which is worse than the retry loop: nothing would ever have told the user, and nothing would have broken until Lufop chose to start enforcing.

`format=meta` is the one call that does check:

```
?key=definitely-not-a-valid-key&format=meta
  → 401 {"error":"unauthorized","message":"Une clé API valide est requise pour accéder à ?format=meta."}
```

and per Lufop's documentation it costs no quota. So: a `LufopAuthError` for 401/403, `async_check_key()` calling meta once per setup, `ConfigEntryAuthFailed` from a rejection, and `async_step_reauth` / `async_step_reauth_confirm` that offers the new key to meta before saving it — a typo is refused rather than stored. The radar endpoint raises the same on 401/403, for the day it starts checking.

Measured: rejection → entry `setup_error`, reason in plain words, a `reauth_confirm` flow in progress. A wrong key typed into that form, against the live API with no injection, comes back `invalid_auth` and the form stays put.

## Smaller items

- **Unique ids key on `entry_id`**, with no migration: the integration is not in the default list yet, so there is nothing deployed to migrate. The entity ids get tidier as a side effect — `sensor.lufop_paris_total` rather than `sensor.lufop_radar_lufop_radar_paris_total`.
- **The count sensor is a `SensorEntity`** with `native_value` and `_attr_state_class`. On the reason, though: the statistics were being kept. `sensor.…_paris_total` has 164 hourly rows going back to 23 August, all recorded while `state_class` was only an attribute — the recorder reads it off the state, not off the class. Changed regardless, because the old way is undocumented and the kind of thing that breaks quietly.
- **`LufopOptionsFlow` calls `super().__init__()`.** I left `OptionsFlowWithConfigEntry` in place: `OptionsFlow.config_entry` only arrived in 2024.11, so moving off it would push the floor past the 2024.7 that the `section` import actually needs.
- **`fr.json`** has `already_configured`; all three files carry 122 identical keys now.
- **The speed value** goes through `saxutils.escape` before it is interpolated.

## `brand/icon.png` — I would like to keep this one

On 2026.x Home Assistant serves that file as the integration's logo:

```
/api/brands/integration/lufop_radar/dark_icon@2x.png → 200, image/png, 25656 bytes
```

which is byte for byte `brand/icon.png`. Removing it takes the logo off the integrations page until a PR to home-assistant/brands lands. I would rather do the brands PR first and drop the file after it is merged — happy to do it the other way round if you prefer.

Thanks again for the time you put into this.

PR #9502, opened 2026-07-26
Adds new integration [berkansezer77/sentinel-alarm]

berkansezer77:
## Checklist

## Links

Link to current release: <https://github.com/berkansezer77/sentinel-alarm/releases/tag/v1.6.2>
Link to successful HACS action (without the `ignore` key): <https://github.com/berkansezer77/sentinel-alarm/actions/runs/30188394303>
Link to successful hassfest action (if integration): <https://github.com/berkansezer77/sentinel-alarm/actions/runs/30188394303>


frenck (changes requested):
Hi @berkansezer77, thanks for the submission! 👋

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

frenck (changes requested):
Hi @berkansezer77, thanks for merging master in back on the 6th. Here is the proper pass.

First, so it does not worry you: the red checks on this PR are not yours. Run `31117979409` failed at "Set up job" with `Failed to resolve action download info. Error: Service Unavailable`, so Hassfest and the JSON schema check never actually executed. Everything that did run passed, and your own repository's validation is green on the `v1.6.2` tag commit. That just needs a rerun on our side.

Reviewed against `v1.6.2`. The declared floor of `2024.11.0` is exactly right: `AlarmControlPanelState` and the `alarm_state` property are absent from core at 2024.10.0 and present at 2024.11.0, so you identified the binding constraint correctly. Three entities documented and three shipped, one service documented and one shipped with the field list and the `mode` options matching the schema exactly. And the auth posture is better than most: all five views are `requires_auth = True` *and* separately gated on `request["hass_user"].is_admin`, the panel is registered `require_admin=True`, and the free-form service step passes the caller's context so Home Assistant's own permission check still applies. The exponential lockout on wrong disarm codes, with a critical notification, is a nice touch.

Four things.

## The integration writes to and deletes from the Lovelace resource store

`__init__.py` lines 406 to 481. It reaches into `hass.data.get("lovelace")`, mutates `resources.loaded = True` after calling `async_load()` directly, then calls `async_create_item` at line 441, `async_update_item` at line 447, and `async_delete_item` at line 450, and retries the whole thing on a widening schedule out to two minutes.

Two problems with that, separate from the general point that an integration should not be editing the user's resource list.

Line 450 deletes every `sentinel-alarm-card-*.js` resource entry that is not the one it wants. That is a delete of an entry the integration did not necessarily create, made without asking.

And `async_unload_entry` at lines 511 to 513 removes none of it. Uninstall the integration and the resource stays, pointing at a URL that no longer serves anything.

You did get the YAML-mode guard right, and your comment explains why you avoided `add_extra_js_url`, which is a real consideration. The answer is to ship the card as a separate `plugin` submission, where HACS writes that resource as the package manager acting for the user and the user knows it happened.

## The stored disarm code is used to bypass its own check

`engine.py` line 1673:

```python
await ent.async_alarm_disarm(self.config.get("code") or None)
```

reached from auto-arrive (line 1759), the clock schedule (line 1702), entity triggers (lines 1732 and 1739) and the push notification Disarm button (line 1852).

Your own comments at `__init__.py` lines 188 to 190 and `alarm_control_panel.py` lines 335 to 338 explain exactly why passing the stored code would make code protection meaningless, and then this path does it. Anyone who can trigger one of those automations, or tap the Disarm action on a notification, disarms without knowing the code. For an alarm integration that is a design decision that at minimum needs to be in the README, and probably needs to be opt-in.

## `async_unload_entry` leaks on every reload

Same lines 511 to 513. It unloads platforms and returns. It never pops `hass.data[DOMAIN]`, never removes the `sentinel_alarm.arm` service, never calls `engine.stop_auto()`, and never unsubscribes the `mobile_app_notification_action` bus listener registered at `engine.py` line 1859.

That listener is added once per setup with no unsub stored, so every reload stacks another handler, and each one of them can disarm the alarm. On an alarm integration that is the one thing you do not want accumulating.

## A 25 MB blocking read on the event loop

`__init__.py` line 314, inside `async def post`:

```python
payload = field.file.read(MAX_UPLOAD + 1)
```

aiohttp spools multipart bodies above a threshold to a temporary file, so that can be a synchronous 25 MB disk read on the loop. It stands out because the write path right below it, `_save`, is correctly dispatched through `async_add_executor_job` at lines 320 to 333, so this looks like one that got missed rather than a pattern.

## Smaller items

- `sentinel-alarm-panel-1.15.0.js` line 4986 interpolates `note` into `el(...)`, whose third argument goes to `innerHTML` at line 9. `note` comes from `this._name(draft.telegram_entity)`, and `_name()` at lines 922 to 925 returns `friendly_name` unescaped. You defined `esc()` at line 17 and use it 216 times, so this is one missed call rather than a habit, and the victim is always an admin, but it is worth closing.
- README lines 97 and 98 say night lights are restored to full brightness on disarm. `alarm_control_panel.py` line 436 calls `stop_night_lights()`, which at `engine.py` lines 1881 to 1892 calls `light.turn_off`. There is no brightness restore anywhere.
- The `sentinel_alarm.arm` service is not mentioned in the README at all, and the card's `background` option is not documented either.
- The v1.6.2 notes say the bundled `brand/` PNGs make Home Assistant show the icon. Nothing reads that directory; Home Assistant sources brand images from the brands repository, and only reads an in-tree `brand/` folder from 2026.3 onward. That is 56 KB of dead payload in the component plus another copy at the repo root.
- `manifest.json` declares `websocket_api` as a dependency and nothing in the tree uses it.
- Unique IDs are constants (`sentinel_alarm_panel`, `sentinel_alarm_tts_openai`, `sentinel_alarm_tts_gemini`) rather than scoped to `entry.entry_id`. Your single-instance guard makes that safe today; it stops being safe the moment multi-instance is added.

The security sweep otherwise came back clean: no secrets anywhere, no eval family, the ffmpeg call is exec-form with the binary from Home Assistant's own manager and audio over stdin, no unpinned or VCS dependencies, no `async_timeout`, and no unparenthesized multi-type `except`.

Fix the Lovelace writes and the unload path, decide what you want the automation disarm to do, and push a release. I will look again.

../Frenck  
<p>
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
  &nbsp;&nbsp;
</p>

berkansezer77:
Thanks for the pass — and for the note on the red checks. That saved me chasing something that was never mine.

Released as [v1.9.0](https://github.com/berkansezer77/sentinel-alarm/releases/tag/v1.9.0).

**The Lovelace resource writes.** Gone entirely — three helpers and the start hook, 95 lines, replaced by nothing. The card now ships from [berkansezer77/sentinel-alarm-card](https://github.com/berkansezer77/sentinel-alarm-card) and HACS registers the resource, which is your point exactly: the package manager acting for the user, with the user knowing. I am submitting it as a plugin separately.

The integration still ships and serves the card file in this release so existing installs keep working — their resource entry is untouched, and both READMEs tell them to delete it before installing the HACS card, since two entries load the file twice and Lovelace then reports the element as missing. I will drop the bundled copy once the plugin lands.

**The stored code.** You were right that the path contradicted the comments a few files over. With a code set, automatic disarming now leaves the alarm armed unless the user turns on an explicit `Let automations disarm` setting, off by default. All four paths you listed funnel through one helper, so it is a single gate.

**`async_unload_entry`.** Now releases the notification listener, the schedule listeners, the `arm` service and the engine. One deviation worth flagging: I pop `engine` rather than the whole `hass.data[DOMAIN]`. The views, the panel and the static paths genuinely outlive the entry, so the `frontend_registered` marker has to survive with them — dropping it would make a reload try to register a static path that is already there.

**The 25 MB read.** Moved to an executor, next to `_save`.

Smaller items: the unescaped `friendly_name`, the unused `websocket_api`, the duplicate `brand/` at the repo root, `sentinel_alarm.arm` and the card background now documented, and the night-lights claim corrected — they are turned off on disarm, not restored to full brightness. That line had been wrong since the feature was written.

Two I did not do, both deliberate:

- **Unique IDs.** Scoping them to `entry.entry_id` now would orphan every existing entity, and with it every user's history, cards and automations. As you say, the single-instance guard makes the constants safe today. I would rather do that together with multi-instance support, where the change actually buys something, than break installs for a rename.
- **`brand/`.** I kept the copy inside the component. By your own reading Home Assistant serves it from 2026.3 onward, so it is live for anyone on a current release rather than dead payload — it is only inert for the older end of the declared range. The duplicate at the repo root was pure waste and is gone. If you would rather it not be carried at all until the floor moves up, say so and I will drop it.

PR #10236, opened 2026-08-22
Deletes integration [lalexdotcom/ha-vacances-fr]

lalexdotcom:
### This is a Delete action

## Links

Link to current release: <https://github.com/lalexdotcom/ha-vacances-fr/releases/tag/1.1.5>
Link to successful HACS action (without the `ignore` key): <https://github.com/lalexdotcom/ha-vacances-fr/actions/runs/27110699755>
Link to successful hassfest action (if integration): <https://github.com/lalexdotcom/ha-vacances-fr/actions/runs/27110699755>

Removing my own integration `lalexdotcom/ha-vacances-fr` from the default catalog.

It is no longer maintained, and it is no longer needed: the same result can be obtained
with the built-in Remote Calendar integration subscribed to the official Education
nationale ICS feeds - one per zone, listed at
https://data.education.gouv.fr/explore/assets/fr-en-calendrier-scolaire/ - plus a few
template sensors.

This PR touches the `integration` file only, per the one-file rule. It first also carried
an entry in `removed` so that people who have it installed would get the repair issue
naming the replacement rather than seeing it vanish silently; that has been dropped to
keep the diff to a single file. Happy to open it as a separate PR, or to leave it to you.

### DIFF
--- a/integration
+++ b/integration
@@ -1667,7 +1667,6 @@
   "lackas/ha-gridx",
   "LaggAt/ha-jokes",
   "LaggAt/hacs-govee",
-  "lalexdotcom/ha-vacances-fr",
   "Lamarqe/ha_openems",
   "lancer73/ha-weather-uploader",
   "larry-wong/bemfa",


lalexdotcom:
Reduced to a single file: the PR now only removes the line from `integration`. The `removed` entry has been dropped from this diff - let me know if you would like it as a separate PR, since it is what would tell people who still have the integration installed where to go instead.

lalexdotcom:
In case #10238 is not the right way to do this and you would rather add the
`removed` entry yourselves, here it is ready to paste, so nobody has to rewrite it:

```json
  {
    "reason": "it is no longer maintained, and the same result can be had with the built-in Remote Calendar integration subscribed to the official Éducation nationale ICS feeds — one per zone, listed at https://data.education.gouv.fr/explore/assets/fr-en-calendrier-scolaire/ — plus a few template sensors",
    "removal_type": "remove",
    "repository": "lalexdotcom/ha-vacances-fr"
  },
```

`remove` rather than `blacklist`: nothing is wrong with the repository, I am
withdrawing it.

The `reason` is deliberately a lower-case clause with no trailing period, because
`async_handle_removed_repositories` interpolates it into the repair issue as
"Because {reason}, `{name}` has been removed from HACS." — so it has to read as
the middle of that sentence. The URL is fenced by spaces on both sides so an
autolinker cannot swallow the comma that follows it.

